PR #19, opened 2020-07-23
Update error message using lowercase b instead

luukverhoeven:
-B is confusing you can't use uppercase b as argument


brandonskerritt:
Thanks! :) @all-contributors please add luukverhoeven for bugs

PR #41, opened 2020-07-23
Fix command typo in README

machawk1:


brandonskerritt:
You're the best! Thanks!!! I would never have seen that

PR #47, opened 2020-07-24
Wait for child nmap process to exit completely

bernardoamc:
In certain OSs a process might become a zombie if we do not explicitlywait for it to finish. This addresses the child not being killed when a parent receives a CTRL+C.

Fixes https://github.com/brandonskerritt/RustScan/issues/14

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -185,10 +185,12 @@ fn main() {
     let vec = command_list.collect::<Vec<&str>>();
 
     // Runs the nmap command and spawns it as a process.
-    Command::new("nmap")
+    let mut child = Command::new("nmap")
         .args(&vec)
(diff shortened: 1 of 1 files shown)


brandonskerritt:
:astonished: Wow! This might just solve one of my biggest headaches so far!!! Let me test it after I get home, if it does -- thank you **SO** much!!!! :D

brandonskerritt:
@all-contributors please add @bernardoamc for code and bugs

bernardoamc:
@brandonskerritt No worries! :) 

I couldn't replicate the `nmap stdout` issue after `wait`ing for the child process on my machine, let's hope this fixes the problem for every OS. <3

PR #52, opened 2020-07-24
Reduce string allocation

bernardoamc:
Most of the time we can just reference the memory location of the string instead of cloning it.


brandonskerritt:
Thanks so much!!! :D

brandonskerritt:
@all-contributors add bernardoamc for design, innovation, bugs, and code

PR #53, opened 2020-07-24
Add -Pn to the Nmap arguments

Isona:
This prevents Nmap from taking extra time doing its initial host detection when we already know ports are up.
Not having this flag occasionally causes hosts to not be scanned as well.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -180,7 +180,7 @@ fn main() {
         exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


brandonskerritt:
Wow! I thought I had done this already, thank you so much for making RustScan better! <3
@all-contributors please add @Isona for bugs

PR #61, opened 2020-07-25
Refactor scanning code into its own module

bernardoamc:
This will give us flexibility to extend this code into multiple scanners in the future and also decouple it from argument parsing and reporting.


bernardoamc:
This is a big change. let me know what you think. :pray:

brandonskerritt:
Looks good to me! :smile:

PR #73, opened 2020-07-26
Use shellwords to parse nmap arguments from user

### LINKED ISSUE #62: Using a parenthesis in script scan flag causes a crash in Nmap
Hello. 

I am attempting to run RustScan against a lab machine using my favorite set of Nmap flags, which includes `--script "(vuln and safe) or default"`; however, this appears to fail if I do it through RustScan, potentially because RustScan misinterprets the parenthesis. See the screenshots below:
## Nmap scan without RustScan:
image

## RustScan on the same box with the same flags
image

Running this in Kali Linux 2020.3. Please let me know if you have any questions.

bernardoamc:
We were not considering user arguments with quotes or other special characters.

Fixes #62 

cc/ @brandonskerritt  We should probably try a few test cases here to see if things will keep working. :smile: 

The command from the issue should be run as:

```sh
$ cargo run -- xx.xx.xxx.xxx -- --script '"(vuln and safe) or default"'
```


brandonskerritt:
I have tried to break this, I could not.

Going to edit the README and help menu to mention this.

bernardoamc:
Nice! Let me adjust #75 with these changes. :)

PR #75, opened 2020-07-26
Refactor code in order to use structopt

bernardoamc:
This allows us to not worry about argument parsing anymore since Rust can automatically infer the parsing logic from the type of our arguments.

**Goals:**

1. Allow us to introduce new arguments in a straightforward manner
2. Not worry about argument parsing


brandonskerritt:
This looks very clean! I love it :D

PR #76, opened 2020-07-26
Format code using 'cargo fmt'

Atul9:


brandonskerritt:
Thanks so much! :D This has been on my TODO list for a while! :D

PR #91, opened 2020-08-01
Fail early when the specified IP format is not valid

bernardoamc:
We can let `structopt` parse the supplied IP for us and fail early when that IP address is invalid. We can also let `IpAddr` let us know when an IP is an ipv4 or ipv6 without relying on the user to supply this information to us.

I've also exracted the logic around building the final `nmap` arguments to its own function in other to make future changes easy to follow.

@brandonskerritt @smackhack could any of you test these changes? :)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,7 +21,8 @@ extern crate log;
 /// specified server may not be able to handle this many socket connections at once.
 struct Opts {
     /// The IP address to scan
-    ip: String,
+    #[structopt(parse(try_from_str))]
+    ip: IpAddr,
 
     ///Quiet mode. Only output the ports. No Nmap. Useful for grep or outputting to a file.
     #[structopt(short, long)]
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -171,22 +157,14 @@ fn main() {
...
-    } else {
-        format!(
-            "{} {} {} {} {} {} {}",
-            &user_nmap_options, "-vvv", "-Pn", "-6", "-p", &ports_str, opts.ip
-        )
-    };
+    let addr = opts.ip.to_string();
+    let user_nmap_args =
+        shell_words::split(&opts.command.join(" ")).expect("failed to parse nmap arguments");
+    let nmap_args = build_nmap_arguments(&addr, &ports_str, &user_nmap_args, opts.ip.is_ipv6());
```
brandonskerritt, line 163: This is some nice code

PR #93, opened 2020-08-01
Added AppDirs

brandonskerritt:
- Added GitHub action to monitor 404 links in README
- Added AppDirs argument

@smackhack @bernardoamc I have an issue where you cannot print appdirs `-a` or `--appdirs` unless an IP is specified. This is because StructOpts requires all arguments to be required. Any suggestions on fixes? :)


bernardoamc:
@brandonskerritt If instead of `bool` you do `Option<bool>` the argument will become optional. :)

src/main.rs
```diff
@@ -19,7 +21,10 @@ extern crate log;
 /// Fast Port Scanner built in Rust.
 /// WARNING Do not use this program against sensitive infrastructure since the
 /// specified server may not be able to handle this many socket connections at once.
+/// - Discord https://discord.gg/rAnvBbg
+/// - GitHub https://github.com/RustScan/RustScan
 struct Opts {
+    // TODO default_value is a hack to make it an optional argument
```
bernardoamc, line 27: If we want `ip` as optional we need to do `Option<ip>`, but we have to adjust the code accordingly if we do that. 😄

src/main.rs
```diff
@@ -24,10 +25,12 @@ extern crate log;
 /// Fast Port Scanner built in Rust.
 /// WARNING Do not use this program against sensitive infrastructure since the
 /// specified server may not be able to handle this many socket connections at once.
+/// - Discord https://discord.gg/rAnvBbg
```
smackhack, line 28: This is the incorrect discord link- it has been deleted and invalidated.
Correct link should be https://discord.gg/GFrQsGy to align with the README.md
```suggestion
/// - Discord https://discord.gg/GFrQsGy
```

brandonskerritt: fixed :)

src/main.rs
```diff
@@ -66,6 +73,26 @@ fn main() {
...
+    config_path.push("rustscan");
+    config_path.push("config.toml");
+
+    if opts.appdirs {
+        // prints config file location and exits
+        println!("The config file is expected to be at {:?}", config_path);
+        exit(1);
+    }
+
+    let ip = match opts.ip {
```
bernardoamc, line 91: Since why are failing without the IP, why do we need to make it optional?

brandonskerritt: The `appdirs` argument is for the config file, Specifically, it will point the user to the location of the config file on their system. When a user runs this flag, they are expecting to see the file path of the config file and not scan at the same time.

If we do not make the IP address optional, the user cannot run the `appdirs` argument unless they also scan at the same time :) 

My personal pet peeve with programs & config files is that they never show you _where_ to put the file, as it is different per system. This flag will help the user solve that little problem :)

bernardoamc: Interesting! I think it's a good idea to mention the config directory. :) 
Could we output this path during the program execution and hide it if the user pass the `quiet` flag? This way we wouldn't need one more option for arguments.

This way we achieve:

1. IP is always required
2. The user is informed of the config directory unless they explicitly pass the `quiet` flag.
3. The interface (command arguments) stays the same for now.

bernardoamc: We can even put this inside the `print_opening()` function, which might become the banner + extra info we want to pass to our users place. 😄 What do you think?

brandonskerritt: This sounds good! We could write something like:
> "Checking for config file at {location}"
Let me make this into a GitHub issue :)

PR #107, opened 2020-08-02
Bee more tests

brandonskerritt:
* Added 6 more tests
* All tests fail on Mac OS because Mac OS file limit. Will try to fix :)


brandonskerritt:
Let me fix CI

brandonskerritt:
this is now #108

PR #108, opened 2020-08-03
Bee fix mac os ulimit

brandonskerritt:
* Added more tests
* Got CI to work with tests
* 90% code coverage

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -300,7 +300,7 @@ mod tests {
             Ok(res) => res,
             Err(_) => panic!("Could not parse IP Address"),
         };
-        let scanner = Scanner::new(addr, 1, 65535, 10_000, Duration::from_millis(5), true);
+        let scanner = Scanner::new(addr, 1, 5000, 250, Duration::from_millis(5), true);
         let scan_result = block_on(scanner.run());
         assert_eq!(1, 1);
     }
@@ -310,7 +310,7 @@ mod tests {
             Ok(res) => res,
             Err(_) => panic!("Could not parse IP Address"),
         };
-        let scanner = Scanner::new(addr, 1, 1, 10_000, Duration::from_millis(5), true);
+        let scanner = Scanner::new(addr, 1, 1, 500, Duration::from_millis(5), true);
         let scan_result = block_on(scanner.run());
         assert_eq!(1, 1);
     }
@@ -320,7 +320,7 @@ mod tests {
             Ok(res) => res,
             Err(_) => panic!("Could not parse IP Address"),
         };
-        let scanner = Scanner::new(addr, 10, 1, 10_000, Duration::from_millis(5), true);
+        let scanner = Scanner::new(addr, 10, 1, 500, Duration::from_millis(5), true);
         let scan_result = block_on(scanner.run());
         assert_eq!(1, 1);
     }
@@ -330,7 +330,7 @@ mod tests {
             Ok(res) => res,
             Err(_) => panic!("Could not parse IP Address"),
         };
-        let scanner = Scanner::new(addr, 400, 445, 10_000, Duration::from_millis(1500), true);
+        let scanner = Scanner::new(addr, 400, 445, 500, Duration::from_millis(1500), true);
         let scan_result = block_on(scanner.run());
         assert_eq!(true, scan_result.contains(&443));
     }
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -217,14 +217,18 @@ fn infer_batch_size(opts: &Opts, ulimit: rlimit::rlim) -> u32 {
...
-            // so we half it to prevent any weird errors propping up because of it.
+        if ulimit < AVERAGE_BATCH_SIZE {
+            // ulimit is smaller than aveage batch size
+            // user must have very small ulimit
+            // decrease batch size to half of ulimit
+            info!("Halving batch_size because ulimit is smaller than average batch size");
+            println!("{}", "WARNING. Your open file description limit is smaller than expected. You can increas the ulimit with the '-u' flag like '-u 5000' to get default size. Or, use the Docker image. If you do not increase ulimit your RustScan speeds will be much slower in comparison to a normal ulimit.".red());
             batch_size = ulimit / 2
         } else if ulimit > DEFAULT_FILE_DESCRIPTORS_LIMIT {
+            info!("Batcg size is now average batch size");
```
bernardoamc, line 228: `Batcg` -> `Batch`

src/main.rs
```diff
@@ -217,14 +217,18 @@ fn infer_batch_size(opts: &Opts, ulimit: rlimit::rlim) -> u32 {
...
-        if ulimit > DEFAULT_FILE_DESCRIPTORS_LIMIT && ulimit > AVERAGE_BATCH_SIZE {
-            // if ulimt is more than the default && the average size on Ubuntu
-            // the user has a weird OS with an incredibly small ulimit
-            // so we half it to prevent any weird errors propping up because of it.
+        if ulimit < AVERAGE_BATCH_SIZE {
+            // ulimit is smaller than aveage batch size
+            // user must have very small ulimit
+            // decrease batch size to half of ulimit
+            info!("Halving batch_size because ulimit is smaller than average batch size");
+            println!("{}", "WARNING. Your open file description limit is smaller than expected. You can increas the ulimit with the '-u' flag like '-u 5000' to get default size. Or, use the Docker image. If you do not increase ulimit your RustScan speeds will be much slower in comparison to a normal ulimit.".red());
```
bernardoamc, line 225: `increas` -> `increase`

src/main.rs
```diff
@@ -217,14 +217,18 @@ fn infer_batch_size(opts: &Opts, ulimit: rlimit::rlim) -> u32 {
...
             batch_size = ulimit / 2
         } else if ulimit > DEFAULT_FILE_DESCRIPTORS_LIMIT {
+            info!("Batcg size is now average batch size");
             batch_size = AVERAGE_BATCH_SIZE
         } else {
+            info!("batch size is ulimit - 100");
```
bernardoamc, line 231: Do we need this `info` given that we already have the `println` on line 147?

src/main.rs
```diff
@@ -256,20 +263,149 @@ mod tests {
...
+        let addr = match "8.8.8.8".parse::<IpAddr>() {
+            Ok(res) => res,
+            Err(_) => panic!("Could not parse IP Address"),
+        };
+        let scanner = Scanner::new(addr, 400, 445, 100, Duration::from_millis(1500), true);
+        block_on(scanner.run());
+        assert_eq!(1, 1);
+    }
+    #[test]
+    fn test_mac_ulimit() {
```
bernardoamc, line 324: How do we know this test is being run in a mac environment?

brandonskerritt: Travis CI runs all tests in OSX. This test tests for the specific problem of Mac's having a small ulimit, the test fails if it does not decrease the ulimit (as OSX will panic). 

See this as proof that we are testing on OSX:
https://github.com/RustScan/RustScan/blob/master/.travis.yml#L5

bernardoamc: Ohh, good one, thanks for explaining! 🎉 

Should we name these tests according to what they are trying to do and not the environment itself? For example `test_mac_ulimit` becomes `automatic_ulimit_reduction`

brandonskerritt: Ah, sure! I'm a new compsci graduate and they don't teach you naming things :wink: but I'm learning :stuck_out_tongue: Give me a few minutes and I'll resubmit :)

bernardoamc: You will get the hang of it in no time 😄

smackhack: I'd like to just quickly share this document, :)

https://doc.rust-lang.org/1.0.0/style/

PR #110, opened 2020-08-03
GH-Releases, Deb's, Caching, Cleanup

SanjoAndPanjo:
Fixed the problem with gh-release key getting leaked by replacing it with an env var ( $GH_0AUTH_TOKEN )
Removed windows build support since an OS dependency is breaking it.
Removed nightly build support since rust fmt is not available within nightly.
Added caching to decrease build time.
Experimented around with deb building and deploying, not sure if it will work, may review the change I made to the deb build script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: rust
 os:
   - linux
   - osx
-  # - windows
 
 arch:
   - amd64
@@ -12,6 +11,9 @@ arch:
 services:
   - docker  
 
+# Save up the cargo build cache
+cache: cargo  
+
 before_install:
 - rustup component add rustfmt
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -12,33 +11,34 @@ arch:
...
-#   provider: releases
-#   api_key: "GITHUB OAUTH TOKEN"
-#   file: "FILE TO UPLOAD"
-#   skip_cleanup: true
-#   on:
-#     tags: true
+deploy:
+  provider: releases
+  api_key: $GH_0AUTH_TOKEN
+  file: target/debug/rustscan
```
brandonskerritt, line 37: If this uploads the binary file itself, that doesn't work as it uses external dependencies which aren't added to the binary :)

.travis.yml
```diff
@@ -12,33 +11,34 @@ arch:
...
+  file: debs/*.deb
+  skip_cleanup: true
+  on:
+    tags: true
```
brandonskerritt, line 41: Does this deploy on push to master? We had a discussion and we'd like this workflow:
* push to master
* new github release automatically created (y.y.x where x is +1)
:sparkle:

bernardoamc: > new github release automatically created (y.y.x where x is +1)

The only downside with this approach is that we might create minor versions that are backward incompatible, which according to semantic versioning shouldn't happen. 😄 People expect that version `y.y.5` won't break after they upgrade from `y.y.4` for example.

scripts/build.sh
```diff
@@ -0,0 +1,4 @@
+# Build Script
+cargo build
```
bernardoamc, line 2: If we are building to release we should do a `cargo build --release` which has all sorts of optimizations and will reduce the binary size. :)

.travis.yml
brandonskerritt: What about releasing on minor? `y.4.y`? 
**Advantages**
* BlackArch & maybe some other community packages build from master branch, by using master as our release we can better provide a guarantee to those using master for their packages
* We have pretty nice CI, a lot of unit tests and soon great code coverage. Not to mention every bit of code has to be reviewed, and we run the code before creating a PR to check for ourselves. At the moment we do not have any specific tests only for releases
* It's pretty cool to be able to say "master branch" == (docker, cargo, AUR, .deb releases) and may prevent users from building master themselves if they know our distributions are just as up to date as mater.

**Disadvantages**
* It can get quite messy if we release to Master a lot but do not release a major version, as in the minor version may reach some high numbers.
* We would release on minor pull requests such as spelling mistakes. To counter this, we'd have to merge pull requests together and create a new pull request (or update an existing one)

bernardoamc: The disadvantages you listed will become more apparent as the project grows. Right now we have these pain points because the project is growing fast, but as it stabilizes we will want to control our releases more tightly in order to not break peoples expectations and force then to constantly update their binaries.

brandonskerritt: Fair enough! We'll use tagged releases :)

Issue #34, opened 2020-07-23
Allow for multiple IP scanning


ophirharpaz:
I think it'd be nice to be able to run _RustScan_ on multiple IPs, either through:

- stdin (comma-separated IPs?) 
- a file (as in nmap's `iL` flag).

Thanks for the nice tool! 🙏

brandonskerritt:
Hey!! I totally agree with this. The code is actually there for this, just.... I need to connect it to the arguments list and do some quality of life stuff 😁✨

sahadnk72:
Yeah, also support for CIDR notation would be great. For eg: 192.168.10.12/24

brandonskerritt:
TO-DO:
- [ ] Allow for multiple ranges
Something like....

```
rustscan --ranges 10.10.10.* 10.10.19.*
```

brandonskerritt:
Okay, I have a pretty good idea of _how_ to do this. Comment if you think I can improve.

* We have a queue data structure, and we place each IP address in this queue.
* We scan the IPs one-at-a-time based on this queue (we pop an IP out and scan it)

## Main
In main, specifically argparsing, we will have 3 methods of inputting IPs
* File input using `rustscan -- file.txt`
* Standard input (comma-seperated IPs) that includes wildcards "10.10.10.*" **and** CIDR "10.0.0.0/24"
* Through the config file #71 (this will come much later)

A function will take these inputs, and iterate over the IPs it contains pushing each one to the queue.

What do ya think? :)

royharoush:
I think that if you can match all of Nmap target input possibilities both in file and the command line it will probably be easier for people to migrate from the masscan+nmap to rust-->nmap.

input types can include CIDRS (/24), and dynamic rangers (4-5.54-34.4.0-255), and and more. 
I dont mean to self promote, but I do have this little script that integrates masscan and and nmap together similar to what you do. 
it can found here:
https://gist.github.com/royharoush/4188a9547daf49e7d918ea86f5975ad6

my point stating it is that the only fault of masscan, and by extension my script, is that while it does randomize the hosts in masscan, it will still perform a host scan one at at time. 

I'm not familiar with rust, but I it would seem very straight-forward to perform a Rust fast scan against randomize ip-ports pairs as part of the main function for the rust scan, and then by extension, scan those in nmaps randomized results as well. 

you can see my super inefficient bash interpretation of this requirement here: 
[...]

PR #117, opened 2020-08-04
Support multiple IPs

bernardoamc:
IPs should be provided in a comma separated list.

Right now every IP will be scanned sequentially and NMAP will run before the next IP scan starts. This will change on the long term once we provide randomized scans in order to not overwhelm hosts or trigger detection systems in place.

Part of #34

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,7 +7,7 @@ use colored::*;
 use futures::executor::block_on;
 use rlimit::Resource;
 use rlimit::{getrlimit, setrlimit};
-use std::process::{exit, Command};
+use std::process::Command;
 use std::{net::IpAddr, time::Duration};
 use structopt::StructOpt;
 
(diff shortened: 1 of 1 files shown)


brandonskerritt:
I should mention that Quiet mode only prints out the ports, nothing else :)

bernardoamc:
> The WARNING and batch size and print_opening shouldn't appear in quiet mode either. What's happened there? Is this a master issue?

I think this is a master issue, but I can fix it here. :) 

> I should mention that Quiet mode only prints out the ports, nothing else :)

We also need to print the IP we are scanning since there might be different open ports for different IPs. 

> I just noticed that in the help menu, Discord and GitHub aren't on new lines. Not your fault (actually my fault), but do you mind correcting this in this PR?

👌 Do you mind posting here the format you want there?

> PS: do you mind adding a test for this? Such as rustscan 0.0.0.0,127.0.0.1

Leave it to me. :)

brandonskerritt:
I can't edit the docstrings as to push to this PR requires me to login with username and password (I can't do with 2FA) so please replace the one above `struct Opts {` with:

```
/// Fast Port Scanner built in Rust. {n}
/// WARNING Do not use this program against sensitive infrastructure since the{n}
/// specified server may not be able to handle this many socket connections at once.{n}
/// {n}- Discord https://discord.gg/GFrQsGy
/// {n}- GitHub https://github.com/RustScan/RustScan
```

src/main.rs
```diff
@@ -28,9 +28,9 @@ extern crate log;
 /// - Discord https://discord.gg/GFrQsGy
```
bernardoamc, line 28: The links were here the entire time 😄

PR #118, opened 2020-08-04
Feature/create alpine based dockerfile

Hydragyrum:
Here's a simple multi-stage docker file with the bare minimum. 

It probably needs a bit of tweaking and testing, though it seems to find things locally :)


brandonskerritt:
Should we replace the default Docker image of RustScan with this one? :)

Hydragyrum:
Might want to run as a normal user rather than root to avoid surprises

CMNatic:
Dokcerbuild file fails locally, not quite sure how the rust base imagine works which is why I avoided it

Pathing / env issues? @Hydragyrum 
image

Hydragyrum:
You'll probably want to configure Travis to push an official image to dockerhub

CMNatic:
The current debian file can be greatly reduced I agree, this was just a "bandaid" until I got the chance (and wanted to use RustScan ASAP without the loopholes of install rust & cargo etc)

Hydragyrum:
Ah you have to checkout the repo to build it, @CMNatic

Hydragyrum:
Added a non-privileged user to run the scan as per docker best-practices

CMNatic:
Sorry for only just getting to this. IRL things and whatnot....

Build works a-okay for me 👍🏻  nice touch on the non-root user @Hydragyrum 

image

Going to confirm with the other reviewers that they're happy for an intermediary `Dockerfile` which uses [rust:alpine](https://hub.docker.com/_/rust) as the base to be shipped with the repo. 

cc @brandonskerritt Does the above sound good to you? With this, we can update the readme and allow users to either build from the above `Dockerfile` which uses rust:alpine as a base or pull from the published image we currently have [DockerHub](https://hub.docker.com/r/cmnatic/rustscan) that compiles from source

The long term idea is to publish an alpine-based docker image that compiles all from source

Hydragyrum:
@CMNatic the final image uses just alpine, the rust image is used for only for building

brandonskerritt:
> Sorry for only just getting to this. IRL things and whatnot....
> 
> Build works a-okay for me 👍🏻 nice touch on the non-root user @Hydragyrum
> 
> image
> 
> Going to confirm with the other reviewers that they're happy for an intermediary `Dockerfile` which uses [rust:alpine](https://hub.docker.com/_/rust) as the base to be shipped with the repo.
> 
> cc @brandonskerritt Does the above sound good to you? With this, we can update the readme and allow users to either build from the above `Dockerfile` which uses rust:alpine as a base or pull from the published image we currently have [DockerHub](https://hub.docker.com/r/cmnatic/rustscan) that compiles from source
> 
> The long term idea is to publish an alpine-based docker image that compiles all from source

As long as you're happy thats cool, do you want to add to the README in this PR and then submit an approval & merge? :)

Issue #69, opened 2020-07-25
Implement most used ports feature


bernardoamc:
The top thousand ports covers 99% of the use cases and might be enough for most people doing CTFs or recon in general.

## Idea

Implement something like `--top-ports` akin to `nmap` that scans the top `1000` TCP ports. In the future we could extend it to also include `UDP`. 

## Work involved

1. Get a list like [this one](https://nullsec.us/top-1-000-tcp-and-udp-ports-nmap-default/) and add it as a FILE or as a constant in the repo.
2. Potentially change or extend our current `Scanner` struct.

brandonskerritt:
We could always build a "learning" list. Log the found ports to a file in key:value {Port_num: frequency} style, and only do the top X ports.

It would be cool to do, because with the automatic ulimit scaling we would start to reach a point where we can say RustScan "learns" or adapts to the host system to become the best / fastest port scanner for their needs :)

bernardoamc:
Something that adapts to the user's host is an awesome idea! :) 

It can also learn or allow a config file for default commands in the future, things like:
* Amount of parallel nmap processes (assuming we go through this route)
* Only TCP or UDP
* Saving these scans to a default location and so on.

brandonskerritt:
I have a rough idea of how I want to do this:
* Ship a JSON object in the config file with all 65535 ports in format `{port_num: found_counter}` where `found_counter` is how many times that port number is found across all IPs.
* The top 1k ports object will also look like this, but when used we will just turn it into a list of the keys. If the user wants top 250 ports, they can specify that. Since we are building the list on run-time, this isn't a problem for us.
* After every scan RustScan (but before Nmap, if it's used)
1. **Make a new thread**
2. Send the ports found to another module, top_ports.rs or something similar.
3. This records the ports in the JSON object. Since it's JSON, it'll be incredibly quick O(1).
4. Sort the JSON object, so the top ports are first. Not sure how to do this in JSON, but I know in Python you can sort dictionaries so it should be possible in Rust.
5. Store the new object in the CONFIG object.
6. When the user requests top X ports, we simply count that many key:values from the config object.
7. End thread and do any clean up if needed. Nmap should be running, so this shouldn't take much time. We can print "storing found ports in config object".

# Thoughts
* This way, our top X ports is also the same object as the ports and their access rates. 
* When we build the JSON object initially, we can +1 to all of the ports Nmap has for their top 1k ports. That way, the user will be able to use top 1k ports from starting the program.
[...]

PR #126, opened 2020-08-05
Added Config File & Top X ports

brandonskerritt:
* Added Config File
1. If config file doesn't exist, downloads it
2. parses the config file as a struct

That's it. It was a lot harder to find the location, download the file if it doesn't exist, and parse it into a struct than I thought as there were so many exceptions :sweat_smile: 

To do next:
* Update `opts` with the values from the new struct `config`
* If it can't download the file, don't panic, inform the user and carry on

This is for #69


bernardoamc:
This is super cool. :) 

What do you think of extracting all this config logic into it's own module like we do for scanner and port strategy? We can try to keep the `main.rs` file lean and just care about argument parsing and orchestrating the other modules.

bernardoamc:
Closing in favour of #185, which expands on the idea and solve conflicts. <3

PR #127, opened 2020-08-05
Add random port scanning

bernardoamc:
This is based on the [Linear Congruential Generator algorithm](https://en.wikipedia.org/wiki/Linear_congruential_generator) which guarantees we will iterate through an entire range in a pseudo random manner.

This implementation should be bound in memory by the batch_size specified by the user, which will be in average 1_000.

**What is happening:**

1. We iterate in a pseudo random order through the range
2. For each iteration we use the returned value (port) to create a `socket` for each provided IP and store them
3. Once we have reached the batch size we actually try to connect to those sockets, storing the ones that we could connect to
4. Go back to step 2 once all values within the range are exhausted


src/scanner/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    /// Runs scan_range with chunk sizes
+    /// If you want to run RustScan normally, this is the entry point used
+    /// Returns all open ports as Vec<u16>
+    pub async fn run(&self) -> Vec<SocketAddr> {
+        let range_iterator = RangeIterator::new(self.start.into(), self.end.into());
+        let mut sockets: Vec<SocketAddr> = Vec::new();
+        let mut open_sockets: Vec<SocketAddr> = Vec::new();
+
+        for port in range_iterator {
```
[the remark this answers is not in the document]
brandonskerritt: Ahh never mind, I didn't see the curly braces. My bad!

src/scanner/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+
+        open_sockets
+    }
+
+    /// Given a slice of sockets, scan them all.
+    /// Returns a vector of open sockets.
+    async fn scan_range(&self, sockets: &[SocketAddr]) -> Vec<SocketAddr> {
+        let mut ftrs = FuturesUnordered::new();
+        for socket in sockets {
+            ftrs.push(self.scan_socket(socket));
```
brandonskerritt, line 85: Okay but this can be an .iter maybe? :)

src/scanner/range_iterator.rs
```diff
@@ -0,0 +1,129 @@
+use gcd::Gcd;
```
brandonskerritt, line 1: This is some really good code, and the algorithm seems very cool. I'll definitely write about this -- hopefully Hackernews crowd will pick up on it and we get more stars, I can imagine they'd love it :D

PR #135, opened 2020-08-07
Add port scanning strategy

### LINKED ISSUE #133: Add an argument to not use randomisation of ports
Randomisation might slow down the program, but it's worth it for normal scans.

However, in a CTF it is incredibly unlikely to have firewalls or IDS block any sort of scan.

We should add an argument `--ctf` which alters the way the program behaves, so it changes itself to be better at CTFs (where only speed & accuracy matter, not noisiness).

In the future, we may make more choices which makes us a better port scanner, but not 100% effective in a CTF. This argument will allow the user to choose :) 

Assigning @bernardoamc since I know you're working on this right now :)

bernardoamc:
Users will be able to pick from:

  - serial: in order to have their ports scanned in ascending order
  - random: in order to have their ports scanned randomly

Fixes #133


brandonskerritt:
Do you still want me to review this?

src/scanner/mod.rs
```diff
@@ -1,5 +1,4 @@
-mod range_iterator;
-use range_iterator::RangeIterator;
+use super::PortStrategy;
```
brandonskerritt, line 1: I did not know you can do this, this is very cool

PR #144, opened 2020-08-10
Add Makefile for cross-compiling to Linux & Mac 👷‍♂️

TGotwig:


TGotwig:
Emoji Powaaa~~ 🤩✨🚀

Yeah right! Basically if you just run `make`, then it would run just the first code-block, so the line below `install` 🙂
If you want to change that behavour, you can add something like this to the top of Makefile:
```Makefile
all: install build shasum
.PHONY: all
```
If you just want to run one specifig code-block, then name it after `make`, like `make build` 🤠
Makefiles can also be executed with Windows through [nmake](https://docs.microsoft.com/en-us/cpp/build/reference/nmake-reference?view=vs-2019) 🏳️‍🌈

brandonskerritt:
Should I include this in the README documentation for the install? :))) 

Also, for your excessive use of emojis (which I love) I'll add you to contributors :D 

@all-contributors please add @TGotwig for distributions or packaging (something like that)

brandonskerritt:
@allcontributors add  @TGotwig for distributions and releases and packaging

(its one of those I swear)

brandonskerritt:
@all-contributors please add @TGotwig for packaging for platform

TGotwig:
Oh cool, thank you! 😊
Yeah it would make the readme shorter and reduce the amount of files inside the repository,
code from [entrypoint.sh](https://github.com/RustScan/RustScan/blob/master/rustscan-debbuilder/entrypoint.sh) could be moved into Makefile 🙂

PR #168, opened 2020-08-26
Added Windows support by disabling rlimit for Windows

brandonskerritt:
- Disabled rlimit for Windows, allowing Windows support (ulimit is hard-set to 1000 on Windows)
- Edited README to update help menu
- Added Windows testing in Travis CI


src/main.rs
```diff
@@ -156,7 +156,14 @@ fn main() {
         std::process::exit(1);
     }
 
-    let ulimit: rlimit::rlim = adjust_ulimit_size(&opts);
+    // Rlimit does not support Windows
+    // set to 1000 if Windows is used
+    let ulimit: rlimit::rlim = if !(cfg!(windows)) {
+        adjust_ulimit_size(&opts)
+    } else {
+        1000
```
bernardoamc, line 164: Can we extract this to a constant? Similar to what we are doing for the ports. :)

bernardoamc:
I think we will also need to adjust imports and variable types.

src/main.rs
brandonskerritt: I can make it a constant, not sure on what you mean by extract though?

brandonskerritt:
> I think we will also need to adjust imports and variable types.

Yup! Merely importing the `rlimit` module requires `libc` (if I remember the Travis output) which is Linux only. We'd have to find a way to either:
1. Conditionally import `rlimit`
2. Find a new library that doesn't panic when it can't find `libc`

brandonskerritt:
> Time to release a new version, there are tons of features that were not released yet. tada

Also, I want to add 1 last feature before release is possible.

When we connect to a socket, if it fails I want it to try X times (as defined via options). It should theoretically be a simple `while` loop :smile:

PR #174, opened 2020-08-27
Allow compilation on windows (#171)

### LINKED ISSUE #171: Add support for windows
**Is your feature request related to a problem? Please describe.**
Even if it's slow, it'd be nice to have windows support so that this crate could be included in packages targeting cross-platform support.

The issue currently preventing windows support is the use of [rlimit](https://crates.io/crates/rlimit), which does not support windows. By adding checks for windows systems and hardcoding ulimit, we may temporarily work around this. I believe [documentation exists](https://docs.microsoft.com/en-us/windows/win32/memory/memory-limits-for-windows-releases?redirectedfrom=MSDN) for setting resource limits on windows, however I'm not very familiar with this sort of low-level stuff so I can't say for sure that this is feasible. Even if it is, it may be outside the scope of this project.

**Describe alternatives you've considered**
My ultimate goal is to include RustScan in another Rust project, so while running via Docker would be useful if I was an end-user, it is ultimately too cumbersome to call docker from rust and then run RustScan via docker via rust. It would be much simpler if I could simply include the crate in my dependencies.

**Additional context**
See #168, which should hopefully close this issue.

caass:
Along with @brandonskerritt's work, fixes #171 

* Conditionally import `std::convert::TryInto;`
* Define a windows-specific `adjust_ulimit_size` that just returns 1000
* Don't tell users that their ulimit is low 😅

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -352,7 +352,7 @@ fn infer_batch_size(opts: &Opts, ulimit: u32) -> u16 {
             if !(cfg!(windows)) {
                 warning!("Your file limit is very small, which negatively impacts RustScan's speed. Use the Docker image, or up the Ulimit with '--ulimit 5000'. ");
             } else {
-                warning!("Windows is known to be much slower than scanning on Unix systems. Use the Docker image if you need faster speeds.")
+                warning!("Windows is known to be much slower than scanning on Unix systems. Use the Docker image if you need faster speeds.");
             }
             info!("Halving batch_size because ulimit is smaller than average batch size");
             batch_size = ulimit / 2
--- a/src/main.rs
+++ b/src/main.rs
@@ -351,6 +351,8 @@ fn infer_batch_size(opts: &Opts, ulimit: u32) -> u16 {
             // decrease batch size to half of ulimit
             if !(cfg!(windows)) {
                 warning!("Your file limit is very small, which negatively impacts RustScan's speed. Use the Docker image, or up the Ulimit with '--ulimit 5000'. ");
+            } else {
+                warning!("Windows is known to be much slower than scanning on Unix systems. Use the Docker image if you need faster speeds.")
             }
             info!("Halving batch_size because ulimit is smaller than average batch size");
             batch_size = ulimit / 2
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,9 +16,11 @@ use futures::executor::block_on;
 use rlimit::{getrlimit, setrlimit, Resource};
 
 use std::collections::HashMap;
+#[cfg(not(target_os = "windows"))]
+use std::convert::TryInto;
 use std::process::Command;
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -345,7 +349,11 @@ fn infer_batch_size(opts: &Opts, ulimit: u32) -> u16 {
             // ulimit is smaller than aveage batch size
             // user must have very small ulimit
             // decrease batch size to half of ulimit
-            warning!("Your file limit is very small, which negatively impacts RustScan's speed. Use the Docker image, or up the Ulimit with '--ulimit 5000'. ");
+            if !(cfg!(windows)) {
+                warning!("Your file limit is very small, which negatively impacts RustScan's speed. Use the Docker image, or up the Ulimit with '--ulimit 5000'. ");
+            } else {
+                warning!("Windows is known to be much slower than scanning on Unix systems. Use the Docker image if you need faster speeds.")
```
brandonskerritt: This needs a semi-colon on the end

src/main.rs
```diff
@@ -329,11 +326,18 @@ fn adjust_ulimit_size(opts: &Opts) -> u32 {
     rlim.try_into().unwrap()
 }
 
+// Rlimit does not support Windows
+// set to 1000 if Windows is used
+#[cfg(target_os = "windows")]
```
brandonskerritt, line 331: Not too sure on the idea of having 2 functions with the same name, @bernardoamc what do you think?

src/main.rs
brandonskerritt: also thanks for adding this! :D

brandonskerritt (changes requested):

src/main.rs
caass: I'm not sure what the convention is, it makes me a little itchy as well to have two functions with the same name. I think though that this is the way to keep the compiler from freaking out, as checking `cfg(target_os = "windows")` in a conditional doesn't cause the compiler to ignore the code. At least, as much as I can tell. I agree though it's a little off-putting, I just went with the quick-and-dirty patch that worked for me.

src/main.rs
caass: 🤦‍♀️ that's what i get for using the github editor smh

caass: Yeah of course! Initially I just stripped all the warnings but I felt like there should be *something* indicating a preference for Unix.

brandonskerritt (approved):

brandonskerritt:
@all-contributors please add @caass for code and bugs

src/main.rs
bernardoamc: If I'm not mistaken we can do an `if/else` with `cfg!(target_os = "windows")`, but I don't have a strong opinion regarding functions with the same name but that get stripped out during compilation. :)

PR #176, opened 2020-08-27
README.md tweaks

niklasmohrin:
Hey there,

I am proposing some slight changes to the README.

1. When the `alias` is proposed, I would expect it to be fully correct and not contain any more helpers like `<rustscan arguments here>`. I think these tips are nice when the usage is first shown, but they should not appear here. I think they were added, when the Docker Image moved to `rustscan/rustscan` (see in git blame or [commit](https://github.com/RustScan/RustScan/commit/ed12888e8138e9da0d2511417556f7a43a50a759)).
2. I added the `bash` decorator to all of the markdown code blocks to enable syntax highlighting. Honestly, the result is a bit underwhelming (I expected at least some highlighting for the `docker` command ¯\\_(ツ)_/¯). It adds some minor highlighting to other commands though and it can't be wrong to have these annotations in place.
3. I thought it looked weird how there were two adjacent code blocks, so I moved them into a single one. However, now this block is the only one, where there is a `$` before the command, which is a bit odd (_open for discussion_).

I did not open an issue, because these changes are so minor and discussion could be done here. If you want me to open an issue for tracking, that is fine too. Anyway, these are just suggestions based on personal preference, so I can fully understand, if you disagree. Let me know what you guys think!

---

Thanks for your work on `rustscan`! 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Once done, you will no longer need to re-download the image (except when RustSca
 You will have to run this command every time, so we suggest aliasing it to something memorable.
 
 ```
-alias rustscan='docker run -it --rm --name rustscan rustscan/rustscan:alpine <rustscan arguments here> <ip address to scan>'
+alias rustscan='docker run -it --rm --name rustscan rustscan/rustscan:alpine'
 ```
 
 Then we can:
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ We strongly recommend using the `alpine` tag, as this is the latest major - stab
 
 Simply run this command against the IP you want to target:
 
-```
+```bash
 docker run -it --rm --name rustscan rustscan/rustscan:alpine <rustscan arguments here> <ip address to scan>
 ```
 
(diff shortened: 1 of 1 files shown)


brandonskerritt (approved):

brandonskerritt:
Hey!

Thanks a lot for these! Every little thing helps :D 

@all-contributors please add @niklasmohrin for documentation :sparkles:

@CMNatic Can you take a look at the first point:
> When the alias is proposed, I would expect it to be fully correct and not contain any more helpers like <rustscan arguments here>. I think these tips are nice when the usage is first shown, but they should not appear here. I think they were added, when the Docker Image moved to rustscan/rustscan (see in git blame or commit).

Since you are in charge of Docker? I'm not sure on what the convention is or what not :P 

Otherwise, if all is good -- we can merge :D

brandonskerritt:
I'm not sure why this is happening, but GitHub won't let me update the branch with master :( 
image

niklasmohrin:
That's weird...
I manually rebased, so it should be fine now.

PR #193, opened 2020-09-02
added no nmap mode

bee-san:
Quiet mode does not print out ports as they are found, leaving users to get impatient.

This adds --without-nmap which prints out ports as they are found but does not run nmap.

A user in John Hammond Discord requested this.


bernardoamc:
@bee-san I think we are missing files here. 😄

bee-san:
> @bee-san I think we are missing files here. smile



> @bee-san I think we are missing files here. smile

Fixed! I reset HEAD 1 too many times and accidentally deleted my work... I gotta watch out for that! 

Also, I had to force-push because it wouldn't let me force-with-lease :(

src/input.rs
```diff
@@ -78,6 +78,10 @@ pub struct Opts {
     #[structopt(long)]
     pub accessible: bool,
 
+    /// Turns off Nmap.
+    #[structopt(long)]
+    pub without_nmap: bool,
```
bernardoamc, line 83: What do you think of calling this `no_nmap` and change `ignore_config` to `no_config`?  Maybe it's time to create a pattern for these optout defaults. :) 

We should also add this option to the `Config` struct so people can add it to their config files if they want.

bee-san: Sounds like a good idea to me! :) I'll change it when I wake up :D

bernardoamc:
@bee-san Everything looks good besides the file `src/scanner/main.rs` where there are a few merge conflicts. :)

bernardoamc:
Actually it seems to be an entire new file? 🤔

bee-san:
For some reason, my local branch is perfectly okay (main is okay), it's just this PR isn't okay. maybe if I close and start a new PR....

PR #195, opened 2020-09-05
Check '.config/rustscan/config.toml' for config

buermarc:
When reading the config file, '$CONFIG_DIR/.config/rustscan/config.toml'
should also be checked. If both '~/.rustscan.toml' and
'$CONFIG_DIR/.config/rustscan/config.toml' exist, then later will be
used.


bee-san:
Thanks so much!!!
@all-contributors please add @buermarc for code :D

bee-san (approved):

PR #196, opened 2020-09-07
Bee add without nmap

bee-san:
Quiet mode does not print out ports as they are found, leaving users to get impatient.

This adds --without-nmap which prints out ports as they are found but does not run nmap.

A user in John Hammond Discord requested this.

### DIFF
--- a/src/scanner/main.rs
+++ b/src/scanner/main.rs
@@ -43,7 +43,6 @@ fn main() {
     let mut opts: Opts = Opts::read();
     let config = Config::read();
     opts.merge(&config);
-
     dbg!(&opts);
 
     info!("Main() `opts` arguments are {:?}", opts);
@@ -282,6 +281,7 @@ mod tests {
             scan_order: ScanOrder::Serial,
             no_nmap: false,
             no_config: false,
+            ignore_config: false,
         };
         let batch_size = infer_batch_size(&opts, 120);
 
@@ -303,6 +303,7 @@ mod tests {
             scan_order: ScanOrder::Serial,
             no_nmap: false,
             no_config: false,
+            ignore_config: false,
         };
         let batch_size = infer_batch_size(&opts, 9_000);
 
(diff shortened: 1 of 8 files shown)


src/scanner/main.rs
```diff
@@ -0,0 +1,475 @@
...
+            ulimit: Some(2_000),
+            command: Vec::new(),
+            accessible: false,
+            scan_order: ScanOrder::Serial,
+<<<<<<< HEAD:src/main.rs
+            no_nmap: false,
+            no_config: false,
+=======
+            ignore_config: false,
+>>>>>>> master:src/scanner/main.rs
```
bee-san, line 287: I don't have this in my local branch....

src/scanner/main.rs
```diff
@@ -0,0 +1,476 @@
...
+mod scanner;
+use scanner::Scanner;
+
+mod port_strategy;
+use port_strategy::PortStrategy;
+
+use colorful::Color;
+use colorful::Colorful;
+use futures::executor::block_on;
+use rlimit::Resource;
```
bee-san, line 17: This is kind of annoying that this entire file is flagged as "new", no idea how to fix it :(

bee-san:
I'm going to have to squash the commits via GitHub when it merges, it's very hard for me to squash them for some reason. Following this:
https://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed
(not rebase but reset, the 2nd answer) causes it to reset to a non-working version which I have to fix again :(

src/scanner/main.rs
```diff
@@ -0,0 +1,476 @@
+extern crate shell_words;
```
bee-san, line 1: It's annoying that this entire file is classes as "new" but I don't know how to fix it :cry:

bernardoamc: I can fix it for you, no worries! :)

bernardoamc: Your local master is probably out of sync with the remote master.

PR #197, opened 2020-09-07
Introduce the -no-nmap option

bernardoamc:
As the name implies this will allow RustScan to to run the scan without starting an nmap scan with ports that are found.

This is basically a copy of: #196


bernardoamc:
@bee-san I had to close your PR. Check if you local master branch is synchronized with the remote one. Let me know if you need help. :)

bee-san (approved):

PR #203, opened 2020-09-10
Small improvement in the scanning engine.

bergabman:


src/scanner/mod.rs
```diff
@@ -48,35 +49,40 @@ impl Scanner {
     /// Returns all open ports as Vec<u16>
     pub async fn run(&self) -> Vec<SocketAddr> {
         let ports: Vec<u16> = self.port_strategy.order();
-        let batch_per_ip: usize = self.batch_size as usize / self.ips.len();
+        // let batch_per_ip: usize = self.batch_size as usize / self.ips.len();
```
bee-san, line 52: Delete code that is not needed instead of commenting it out :)

src/scanner/mod.rs
```diff
@@ -48,35 +49,40 @@ impl Scanner {
...
-        }
-
-        open_sockets
-    }
-
-    /// Given a slice of sockets, scan them all.
-    /// Returns a vector of open sockets.
-    async fn scan_ports(&self, ports: &[u16]) -> Vec<SocketAddr> {
+        let mut targets: VecDeque<SocketAddr> =
+            VecDeque::with_capacity(self.ips.len() * ports.len());
```
bee-san, line 55: What does this code do? :)

src/scanner/mod.rs
```diff
@@ -48,35 +49,40 @@ impl Scanner {
...
+                targets.push_back(SocketAddr::new(*ip, port));
             }
         }
 
-        let mut open_sockets: Vec<SocketAddr> = Vec::new();
-        while let Some(result) = ftrs.next().await {
-            match result {
-                Ok(socket) => open_sockets.push(socket),
-                _ => {}
+        while ftrs.len() < self.batch_size as usize {
```
bee-san, line 64: should be <= i think

src/scanner/mod.rs
```diff
@@ -48,35 +49,40 @@ impl Scanner {
...
         }
 
+        loop {
+            match ftrs.next().await {
+                Some(result) => {
+                    match result {
+                        Ok(socket) => open_sockets.push(socket),
+                        Err(_) => {}
+                    }
+                    if !targets.is_empty() {
```
bee-san, line 77: This part does the scanning? :)

bee-san (changes requested):

src/scanner/mod.rs
bergabman: Creates a VecDeque, a double ended queue with capacity. Take the number of ips and ports, multiply them and we have the amount of ip:port combinations go into the targets vector. Reason to use with_capacity is to avoid memory reallocation when the vector have to grow.

src/scanner/mod.rs
bergabman: Correct, we miss one future there.

src/scanner/mod.rs
bergabman: Yes, checking on available results of futures, and when one is finished, a new one gets pushed to the stream as FuturesUnordered allows adding new futures while it's running. In case of success/open port, the socket gets added to the returnvector.

src/scanner/mod.rs
bergabman: It's gone.

src/scanner/mod.rs
bergabman: Fixed.

bee-san:
@all-contributors please add @bergabman for code, bugs, design

bee-san (approved):

PR #208, opened 2020-09-12
Introduce SocketIterator

bernardoamc:
The goal of this iterator is to generate sockets based on combinations of IPs and ports in a memory-bounded way, we only consume the next item of this iterator when a socket has been processed by our scanner.

Before this change we were generating every combination of IP and port and storing it in a VecDeque, and while this is fine for a small subset of IPs and ports it can quickly get out of hand with multiple IPs and a big port range.

### DIFF
--- a/src/scanner/mod.rs
+++ b/src/scanner/mod.rs
@@ -1,12 +1,14 @@
 use super::PortStrategy;
 
+mod socket_iterator;
+use socket_iterator::SocketIterator;
+
 use async_std::io;
 use async_std::net::TcpStream;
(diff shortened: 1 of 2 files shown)


bernardoamc:
@bergabman check this out if you have time, you know this code pretty well by now and it's great to have multiple reviewers. 😄

bee-san (approved):

bee-san:
Beautiful code! :D

bergabman:
Awesome code. Keeping memory usage as low as possible.

PR #217, opened 2020-09-14
fixed debug log

bee-san:
Someone left a debug statement in their code, this fixes it :)


bee-san:
Force pushing this without code review as I just removed the debug print statement.

PR #221, opened 2020-09-14
Fix behaviour of --no-nmap flag

### LINKED ISSUE #202: Disable nmap option triggers nmap anyway
**Describe the bug**
When running the latest version `rustscan 1.8.0`, the flag for disabling nmap transition seems to be ignored.

**To Reproduce**

1. `docker run -it --rm --name rustscan rustscan/rustscan:latest 192.168.1.1/24 -t 500 -b 5000 -r 1-1000 --no-nmap`
2. Observe the `Starting Nmap 7.80 ( https://nmap.org ) at 2020-09-10 04:14 UTC` following the open port report.

**Expected behavior**
I think the use of this flag should stop further execution of nmap and just report the output.

**Desktop (please complete the following information):**
 - OS: Docker on MacOS

### LINKED ISSUE #214: Include IP and Port when using --quiet mode
When running a port scan and using the `--quiet` option, the program returns only the open port number, which makes it impossible to correlate this to the source host/ip.  Please include the source IP in the returned output.

Example
```
[src/main.rs:45] &opts = Opts {
    addresses: [
        "10.0.0.0/24",
    ],
    ports: Some(
        [
            21,
            22,
            23,
            161,
        ],
    ),
    range: None,
    no_config: false,
    quiet: true,
    accessible: false,
    no_nmap: false,
    batch_size: 750,
    timeout: 2000,
    ulimit: None,
    scan_order: Serial,
    command: [],
}
22
22
22
22
22
22
22
22
22
22
22
22
22
22
22
22
22
23
22,23
22
22
22
```

bernardoamc:
At some point we forgot to check for this flag before actually running our post scanning tasks. I've also adjusted some of our messages for reporting.

Closes #214
Closes #202

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -85,13 +85,11 @@ fn main() {
         // If we got here it means the IP was not found within the HashMap, this
         // means the scan couldn't find any open ports for it.
 
-        let x = format!("{} Looks like I didn't find any open ports for {:?}. This is usually caused by a high batch size.
-        \n*I used {} batch size, consider lowering to {} with {} or a comfortable number for your system.
+        let x = format!("Looks like I didn't find any open ports for {:?}. This is usually caused by a high batch size.
+        \n*I used {} batch size, consider lowering it with {} or a comfortable number for your system.
         \n Alternatively, increase the timeout if your ping is high. Rustscan -t 2000 for 2000 milliseconds (2s) timeout.\n",
-        "ERROR",
         ip,
         opts.batch_size,
-        (opts.batch_size / 2).to_string(),
         "'rustscan -b <batch_size> <ip address>'");
         warning!(x, opts.quiet);
     }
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -85,28 +85,26 @@ fn main() {
...
-        (opts.batch_size / 2).to_string(),
         "'rustscan -b <batch_size> <ip address>'");
         warning!(x, opts.quiet);
     }
 
     for (ip, ports) in ports_per_ip.iter_mut() {
         let nmap_str_ports: Vec<String> = ports.into_iter().map(|port| port.to_string()).collect();
 
-        detail!("Starting Nmap", opts.quiet);
+        detail!("Starting Nmap", opts.quiet || opts.no_nmap);
```
bee-san, line 100: WOAH! I wrote this macro and I never realised you could just pas in bools like that. :exploding_head:

bee-san (approved):

Issue #187, opened 2020-08-31
Allow use of top ports in RustScan


bee-san:
Right now we can:
* Get top ports from the config file
* Download & place config file if it doesn't already exist

Ideally, this is what should happen:
* Add an argument to RustScan to specify top ports is to be used.
* Use top ports (a list of individual ports) as the port input.

PR #230, opened 2020-09-18
Added top ports, fixed accessibility.

bergabman:
Fixed issue #187

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -50,14 +50,18 @@ fn main() {
 
     debug!("Main() `opts` arguments are {:?}", opts);
 
-    if !opts.quiet && !opts.accessible {
-        print_opening();
+    if !opts.greppable && !opts.accessible {
+        print_opening(&opts);
     }
 
     let ips: Vec<IpAddr> = parse_addresses(&opts);
 
     if ips.is_empty() {
-        warning!("No IPs could be resolved, aborting scan.", false);
+        warning!(
+            "No IPs could be resolved, aborting scan.",
+            opts.greppable,
+            opts.accessible
+        );
         std::process::exit(1);
     }
 
@@ -68,8 +72,9 @@ fn main() {
         &ips,
         batch_size,
         Duration::from_millis(opts.timeout.into()),
-        opts.quiet,
+        opts.greppable,
         PortStrategy::pick(opts.range, opts.ports, opts.scan_order),
+        opts.accessible,
     );
     debug!("Scanner finished building: {:?}", scanner);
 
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -163,18 +163,22 @@ Faster Nmap scanning with Rust."#;
...
     funny_opening!();
 
-    let config_path = match dirs::config_dir() {
-        Some(mut path) => {
-            path.push("rustscan");
-            path.push("config.toml");
-            path
-        }
-        None => panic!("Couldn't find config dir."),
+    let mut config_dir = match dirs::config_dir() {
```
bee-san, line 166: we only use home dir now, config + home are too complex. I probably forgot to remove this from my code, but Kaeros and I agreed :laughing:

src/main.rs
```diff
@@ -163,18 +163,22 @@ Faster Nmap scanning with Rust."#;
...
+    config_dir.push("rustscan");
+    config_dir.push("config.toml");
+
+    let mut home_dir = match dirs::home_dir() {
+        Some(dir) => dir,
+        None => panic!("Could not infer config file path."),
     };
+    home_dir.push(".rustscan.toml");
 
     detail!(format!(
```
bee-san, line 179: This is a bug I made, pretty pleasee can you fix it :pleading_face: 
This doesn't work with grep mode or accessible mode :(
Can you uhhhh get the config args (or config file struct) and fix my bug? 
See this code for the syntax.
https://github.com/RustScan/RustScan/blob/fa072bac1235fbbfe615893f56f992c4ae7b0cb0/src/tui.rs
detail!("message", greppable, accessible);

src/main.rs
```diff
@@ -481,6 +493,7 @@ mod tests {
             scan_order: ScanOrder::Serial,
             no_config: false,
             no_nmap: false,
+            top: false,
```
bee-san, line 496: Can you add another one of these tests and make `top: true`? I'm not sure if it'll work with our CI :shrug: but if it does that's great and we can maintain high test coverage :smile:

bee-san (changes requested):

src/main.rs
bee-san: oh i clicked the wrong thing. I meant the `detail!` you can see, but I clicked the `home_dir.push` :laughing:

bee-san: PS: if you do this we can officially say we are screen reader friendly which is something not even Nmap can say! :smile:

src/main.rs
bee-san: on 2nd thoughts this would not work right now

src/main.rs
bergabman: All right, removed from the places where it's checked.

src/main.rs
bergabman: Changed the `print_opening()` function to accept the opts struct, which now contains both the greppable and accessible options, so we can match on those with the `detail!()` macro. Now works fine in both cases.

src/main.rs
bergabman: I have a plan to refactor the tests a little bit, but that will be better in another PR, because most of them have to change a little. Then we can make a new test for this functionality as well.

bee-san (approved):

Issue #81, opened 2020-07-28
RustScan Scripting Engine


bee-san:
# TL;DR
* Write exploit scripts in Python. Users can do whatever they want. Get robots.txt, check if FTP is anonymous.
* Use a custom tagging system (set theory based and argument based) to allow the user to create their own perfect categories of scripts.

# Problem
Users want to write exploit modules with RustScan. Specifically, they feel ike Nmap's LUA isn't modern enough, or they want to pipe the output of RustScan into the input of a script.

They want to write exploit scripts using RustScan.

But we, the core maintainers of RustScan, cannot possibly compete with Nmap for scripts -- especially when we are focussed mostly on improving RustScan.

# Solution
We introduce a RustScan Scripting Engine which will allow anyone to create scripts, or custom exploits and use them in RustScan. 

This way, the community can build scripts and exploits if they want -- while we are RustScan work on improving the scanner generally.

## Python
We should use Python to allow users to write scripts. Rust. Firstly, Rust has always had very good support for foreign function interfaces (FFI) with Python. You can even write [inline Python with Rust.](https://docs.rs/inline-python/0.5.3/inline_python/)

We should consider using Pyo3. 
[https://github.com/PyO3/pyo3](https://github.com/PyO3/pyo3).

Python is easily the most friendly programming language around, Many people know it, and a lot of exploits (from Exploit-DB) are written in Python.

Giving the user the ability to not only write custom scripts for RustScan, but to do so in Python would be a huge plus.

## Default template
We need to create a default template, so RustScan can easily intrepret the template and use it.

The exact template is unknown, until we look at Pyo3. However, Pyo3 has this Rust code (for calling Python from Rust):

```rust
use pyo3::prelude::*;
use pyo3::types::IntoPyDict;

fn main() -> Result<(), ()> {
    Python::with_gil(|py| {
        main_(py).map_err(|e| {
          // We can't display Python exceptions via std::fmt::Display,
          // so print the error here manually.
          e.print_and_set_sys_last_vars(py);
        })
    })
}

fn main_(py: Python) -> PyResult<()> {
    let sys = py.import("sys")?;
    let version: String = sys.get("version")?.extract()?;
    let locals = [("os", py.import("os")?)].into_py_dict(py);
    let code = "os.getenv('USER') or os.getenv('USERNAME') or 'Unknown'";
    let user: String = py.eval(code, None, Some(&locals))?.extract()?;
    println!("Hello {}, I'm Python {}", user, version);
[...]

PR #242, opened 2020-09-23
Scripting Engine Implementation

bergabman:
I started to implement the scripting engine how I thought it could work smooth. We can make this structure dynamic enugh to fill in all the necessary info we talked about before, like `core_approved` and description of the script. My idea would be to make the interpreted language files in a self contained way, which means the first 4-5 lines of the file could contain the tags in comment space for the interpreter, and the rest of the file can be the script itself.
I'm not sure how we should handle binary files, if the user want's to run a binary we have to store the tags on a different location.

For a main rustscan_scripting config file we could use toml format, where we store some basic variables in case we have any. To read the scripting config, we could add a new arg like `--scripting`

```
tags = ["core_approved", ""]


# optional field, if it's given then only those scripts will run which has a tag ports = [80]
# 
# ex.: 
# ports = [80] 
# ports = [80,81,8080]
ports = [80]


# optional field, if it's given, only that(those) interpreter(s) will be used 
# 
interpreters = ["python3", "bash"]
```

Please check it out how it looks and give comments how to imporve, or if it's even a good idea to do it this way. @bee-san @bernardoamc @CMNatic

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,9 @@ serde_derive = "1.0.116"
 cidr-utils = "0.5.0"
 itertools = "0.9.0"
 trust-dns-resolver = { version = "0.19.5", features = ["dns-over-rustls"] }
+anyhow = "1.0.33"
+subprocess = "0.2.6"
+text_placeholder = { version = "0.4", features = ["struct_context"] }
 
 [dev-dependencies]
 wait-timeout = "0.2"
--- a/fixtures/test_rustscan_scripts.toml
+++ b/fixtures/test_rustscan_scripts.toml
@@ -1,0 +1,14 @@
+# Test/Example ScriptConfig file
+
+# Tags to filter on scripts. Only scripts containing all these tags will run.
+tags = ["core_approved", "example"]
+
+# If it's present then only those scripts will run which has a tag ports = "80". Not yet implemented.
+#
+# ex.:
+# ports = [80]
+# ports = [80,81,8080]
+ports = [80]
+
+# Only this developer(s) scripts to run. Not yet implemented.
+developer = ["example"]
--- a/fixtures/test_script.py
+++ b/fixtures/test_script.py
@@ -1,0 +1,13 @@
+#!/usr/bin/python3
+#tags = ["core_approved", "example",]
+#developer = [ "example", "https://example.org" ]
+#trigger_port = "80"
+#call_format = "python3 {{script}} {{ip}} {{port}}"
+
+# Sriptfile parser stops at the first blank line with parsing.
+# This script will run itself as an argument with the system installed python interpreter, only scanning port 80.
+# Unused filed: ports_separator = ","
+
+import sys
+
+print('Python script ran with arguments', str(sys.argv))
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,7 +1,6 @@
 use serde_derive::Deserialize;
 use std::collections::HashMap;
 use std::fs;
-
 use structopt::{clap::arg_enum, StructOpt};
 
 const LOWEST_PORT_NUMBER: u16 = 1;
@@ -15,6 +14,19 @@ arg_enum! {
     pub enum ScanOrder {
         Serial,
         Random,
+    }
+}
+
+arg_enum! {
+    /// Represents the scripts variant.
+    ///   - none will avoid running any script, only portscan results will be shown.
+    ///   - default will run the default embedded nmap script, that's part of RustScan since the beginning.
+    ///   - custom will read the ScriptConfig file and the available scripts in the predefined folders
+    #[derive(Deserialize, Debug, StructOpt, Clone, PartialEq, Copy)]
+    pub enum ScriptsRequired {
+        None,
+        Default,
+        Custom,
     }
 }
 
@@ -49,7 +61,7 @@ fn parse_range(input: &str) -> Result<PortRange, String> {
     }
 }
 
-#[derive(StructOpt, Debug)]
+#[derive(StructOpt, Debug, Clone)]
 #[structopt(name = "rustscan", setting = structopt::clap::AppSettings::TrailingVarArg)]
 /// Fast Port Scanner built in Rust.
 /// WARNING Do not use this program against sensitive infrastructure since the
@@ -81,10 +93,6 @@ pub struct Opts {
     #[structopt(long)]
     pub accessible: bool,
 
-    /// Turns off Nmap.
-    #[structopt(long)]
-    pub no_nmap: bool,
-
     /// The batch size for port scanning, it increases or slows the speed of
     /// scanning. Depends on the open file limit of your OS.  If you do 65535
     /// it will do every port at the same time. Although, your OS may not
@@ -111,17 +119,21 @@ pub struct Opts {
     #[structopt(long, possible_values = &ScanOrder::variants(), case_insensitive = true, default_value = "serial")]
     pub scan_order: ScanOrder,
 
-    /// The Nmap arguments to run.
+    /// Level of scripting required for the run.
+    #[structopt(long, possible_values = &ScriptsRequired::variants(), case_insensitive = true, default_value = "default")]
+    pub scripts: ScriptsRequired,
+
+    /// Use the top 1000 ports.
+    #[structopt(long)]
+    pub top: bool,
+
+    /// The Script arguments to run.
     /// To use the argument -A, end RustScan's args with '-- -A'.
     /// Example: 'rustscan -T 1500 127.0.0.1 -- -A -sC'.
     /// This command adds -Pn -vvv -p $PORTS automatically to nmap.
     /// For things like --script '(safe and vuln)' enclose it in quotations marks \"'(safe and vuln)'\"")
     #[structopt(last = true)]
     pub command: Vec<String>,
-
-    /// Use the top 1000 ports.
-    #[structopt(long)]
-    pub top: bool,
 }
 
 #[cfg(not(tarpaulin_include))]
@@ -160,7 +172,8 @@ impl Opts {
         }
 
         merge_required!(
-            addresses, greppable, accessible, batch_size, timeout, tries, scan_order, command
+            addresses, greppable, accessible, batch_size, timeout, tries, scan_order, scripts,
+            command
         );
     }
 
@@ -204,10 +217,10 @@ pub struct Config {
     batch_size: Option<u16>,
     timeout: Option<u32>,
     tries: Option<u8>,
-    no_nmap: Option<bool>,
     ulimit: Option<rlimit::RawRlim>,
     scan_order: Option<ScanOrder>,
     command: Option<Vec<String>>,
+    scripts: Option<ScriptsRequired>,
 }
 
 #[cfg(not(tarpaulin_include))]
@@ -251,7 +264,7 @@ impl Config {
 
 #[cfg(test)]
 mod tests {
-    use super::{Config, Opts, PortRange, ScanOrder};
+    use super::{Config, Opts, PortRange, ScanOrder, ScriptsRequired};
     impl Config {
         fn default() -> Self {
             Self {
@@ -263,10 +276,10 @@ mod tests {
                 timeout: Some(1_000),
                 tries: Some(1),
                 ulimit: None,
-                no_nmap: Some(false),
                 command: Some(vec!["-A".to_owned()]),
                 accessible: Some(true),
                 scan_order: Some(ScanOrder::Random),
+                scripts: None,
             }
         }
     }
@@ -284,10 +297,10 @@ mod tests {
                 ulimit: None,
                 command: vec![],
                 accessible: false,
-                no_nmap: false,
                 scan_order: ScanOrder::Serial,
                 no_config: true,
                 top: false,
+                scripts: ScriptsRequired::Default,
             }
         }
     }
@@ -320,6 +333,7 @@ mod tests {
         assert_eq!(opts.command, config.command.unwrap());
         assert_eq!(opts.accessible, config.accessible.unwrap());
         assert_eq!(opts.scan_order, config.scan_order.unwrap());
+        assert_eq!(opts.scripts, ScriptsRequired::Default)
     }
 
     #[test]
--- a/src/main.rs
+++ b/src/main.rs
@@ -3,7 +3,7 @@ extern crate shell_words;
 mod tui;
 
 mod input;
-use input::{Config, Opts, PortRange, ScanOrder};
+use input::{Config, Opts, PortRange, ScanOrder, ScriptsRequired};
 
 mod scanner;
 use scanner::Scanner;
@@ -14,6 +14,9 @@ use port_strategy::PortStrategy;
 mod benchmark;
 use benchmark::{Benchmark, NamedTimer};
 
+mod scripts;
+use scripts::{init_scripts, Script, ScriptFile};
+
 use cidr_utils::cidr::IpCidr;
 use colorful::{Color, Colorful};
 use futures::executor::block_on;
@@ -23,7 +26,6 @@ use std::fs::File;
 use std::io::{prelude::*, BufReader};
 use std::net::{IpAddr, ToSocketAddrs};
 use std::path::Path;
-use std::process::Command;
 use std::time::Duration;
 use trust_dns_resolver::{config::*, Resolver};
 
@@ -51,6 +53,10 @@ fn main() {
     opts.merge(&config);
 
     debug!("Main() `opts` arguments are {:?}", opts);
+
+    let scripts_to_run: Vec<ScriptFile> =
+        init_scripts(opts.scripts).expect("could not initiate scripting part.");
+    debug!("Scripts initialized {:?}", &scripts_to_run);
 
     if !opts.greppable && !opts.accessible {
         print_opening(&opts);
@@ -112,46 +118,71 @@ fn main() {
         warning!(x, opts.greppable, opts.accessible);
     }
 
-    let mut nmap_bench = NamedTimer::start("Nmap");
+    let mut script_bench = NamedTimer::start("Scripts");
     for (ip, ports) in ports_per_ip.iter_mut() {
-        let nmap_str_ports: Vec<String> = ports.into_iter().map(|port| port.to_string()).collect();
+        let vec_str_ports: Vec<String> = ports.into_iter().map(|port| port.to_string()).collect();
 
         // nmap port style is 80,443. Comma separated with no spaces.
-        let ports_str = nmap_str_ports.join(",");
+        let ports_str = vec_str_ports.join(",");
 
-        // if greppable mode is on nmap should not be spawned
-        if opts.greppable || opts.no_nmap {
+        // if option scripts is none, no script will be spawned
+        if opts.greppable || opts.scripts.clone() == ScriptsRequired::None {
             println!("{} -> [{}]", &ip, ports_str);
             continue;
         }
-        detail!("Starting Nmap", opts.greppable, opts.accessible);
+        detail!("Starting Script(s)", opts.greppable, opts.accessible);
 
-        let addr = ip.to_string();
-        let user_nmap_args =
-            shell_words::split(&opts.command.join(" ")).expect("failed to parse nmap arguments");
-        let nmap_args = build_nmap_arguments(&addr, &ports_str, &user_nmap_args, ip.is_ipv6());
+        // Run all the scripts we found and parsed based on the script config file tags field.
+        for mut script_f in scripts_to_run.clone() {
+            output!(
+                format!("Script to be run {:?}\n", script_f.call_format,),
+                opts.greppable.clone(),
+                opts.accessible.clone()
+            );
 
-        output!(
-            format!(
-                "The Nmap command to be run is nmap {}\n",
-                &nmap_args.join(" ")
-            ),
-            opts.greppable.clone(),
-            opts.accessible.clone()
-        );
+            // This part allows us to add commandline arguments to the Script call_format, appending them to the end of the command.
+            if opts.command.len() > 0 {
+                let user_extra_args: Vec<String> = shell_words::split(&opts.command.join(" "))
+                    .expect("Failed to parse extra user commandline arguments");
+                if script_f.call_format.is_some() {
+                    let mut call_f = script_f.call_format.unwrap();
+                    call_f.push_str(&format!(" {}", &user_extra_args.join(" ")));
+                    script_f.call_format = Some(call_f);
+                }
+            }
 
-        // Runs the nmap command and spawns it as a process.
-        let mut child = Command::new("nmap")
-            .args(&nmap_args)
-            .spawn()
-            .expect("failed to execute nmap process");
-
-        child.wait().expect("failed to wait on nmap process");
+            // Building the script with the arguments from the ScriptFile, and ip-ports.
+            let script = Script::build(
+                script_f.path,
+                *ip,
+                ports.to_vec(),
+                script_f.port,
+                script_f.ports_separator,
+                script_f.tags,
+                script_f.call_format,
+            );
+            match script.run() {
+                Ok(script_result) => {
+                    detail!(
+                        format!("{}", script_result),
+                        opts.greppable,
+                        opts.accessible
+                    );
+                }
+                Err(e) => {
+                    warning!(
+                        &format!("Error {}", e.to_string()),
+                        opts.greppable,
+                        opts.accessible
+                    );
+                }
+            }
+        }
     }
 
     // To use the runtime benchmark, run the process as: RUST_LOG=info ./rustscan
-    nmap_bench.end();
-    benchmarks.push(nmap_bench);
+    script_bench.end();
+    benchmarks.push(script_bench);
     rustscan_bench.end();
     benchmarks.push(rustscan_bench);
     debug!("Benchmarks raw {:?}", benchmarks);
@@ -165,7 +196,7 @@ fn print_opening(opts: &Opts) {
 | {}  }| { } |{ {__ {_   _}{ {__  /  ___} / {} \ |  `| |
 | .-. \| {_} |.-._} } | |  .-._} }\     }/  /\  \| |\  |
 `-' `-'`-----'`----'  `-'  `----'  `---' `-'  `-'`-' `-'
-Faster Nmap scanning with Rust."#;
+The Modern Day Port Scanner."#;
     println!("{}", s.gradient(Color::Green).bold());
     let info = r#"________________________________________
 : https://discord.gg/GFrQsGy           :
@@ -296,27 +327,6 @@ fn read_ips_from_file(
         }
     }
     Ok(ips)
-}
-
-#[cfg(not(tarpaulin_include))]
-fn build_nmap_arguments<'a>(
-    addr: &'a str,
-    ports: &'a str,
-    user_args: &'a Vec<String>,
-    is_ipv6: bool,
-) -> Vec<&'a str> {
-    let mut arguments: Vec<&str> = user_args.iter().map(AsRef::as_ref).collect();
-    arguments.push("-vvv");
-
-    if is_ipv6 {
-        arguments.push("-6");
-    }
-
-    arguments.push("-p");
-    arguments.push(ports);
-    arguments.push(addr);
-
-    arguments
 }
 
 fn adjust_ulimit_size(opts: &Opts) -> RawRlim {
--- a/src/scripts/mod.rs
+++ b/src/scripts/mod.rs
@@ -1,0 +1,355 @@
+//! Scripting engine to run scripts based on tags.
+//! This module serves to filter and run the scripts selected by the user.
+//!
+//! A new commandline and configuration file option was added.
+//!
+//! --scripts
+//!
+//!      default
+//!          This is the default behavior, like as it was from the beginning of RustScan.
+//!          The user do not have to chose anything for this. This is the only script embedded in RustScan running as default.
+//!
+//!      none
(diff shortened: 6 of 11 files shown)


bergabman:
Right now it works exactly the same, nmap runs as before just with the new structure. So this is actually a working version, just for the record. :)

bergabman:
#81 the issue

src/main.rs
```diff
@@ -14,6 +14,9 @@ use port_strategy::PortStrategy;
 mod benchmark;
 use benchmark::{Benchmark, NamedTimer};
 
+mod scripts;
+use scripts::{Executable::Nmap, Script};
```
bee-san, line 18: Would we have to add every core script here?

src/scripts/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+//!
+use anyhow::{anyhow, Result};
+use std::net::IpAddr;
+use std::path::PathBuf;
+use subprocess::{Exec, ExitStatus};
+use which;
+
+// struct ScriptEngine {
+//     nmap: Bin,
+//     python: Bin,
```
bee-san, line 12: We should allow them, in the config file, to specify a file path for Python too :) 

We're going to get a lot of comments like "Python doesn't work" when their default `python` is 2.7 and not `3.7`. Also, we'd need to make sure scripts are compatible with lower version of Python, or to have a version of python mentioned in the script name. 

So if the script tag is `python3.7` it doesn't run if the user has `3.6.9`

src/scripts/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+// impl ScriptEngine {
+//     fn build(&self, interpreter: String, arguments: Vec<String>, ip: String, ports: Vec<u16>) {
+
+//     }
+//     fn execute(script: Script, addr: &str, ports: Vec<_>, arguments: Vec<String>) {
+
+//     }
+// }
+
+pub struct Script {
```
bee-san, line 24: This is good, I can't imagine anything else we'd want to add.

Perhaps a specific argument for the language that will execude it?

```rust
language: String
```

So we can easily see if it's Python, Perl, Shell etc

src/scripts/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+
+            // optional path to the executable
+            path: path,
+
+            // user given arguments on commandline after --
+            arguments: arguments,
+            ip: ip,
+            ports: ports,
+
+            /*  script tags stored in file header? We can use toml to read it into a struct after parsing the headers manually first
```
bee-san, line 69: We should actually do something like this, ideally:

```
#! Shebang thingy here
# RSE START
# Tags: Core, HTTP, Web, HTTPS
# Ports: 80, 443
# RSE END
```

That way we don't have to parse "maybe line 1 is a shebang?" and "does their comments end after 5 lines or 4?"

Also, if we use TOML like structure within them we can easily parse for the struct.

And searching for `RSE START` and `RSE END` are easier.

Use "end" instead of "stop" because "stop" is physically closer to "start" (same first 2 letters, small minor thing but it might help people read the code)

src/scripts/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+                so read file, parse first 5 lines, remove first line (shebang) then throw it to toml to parse the tags to the struct.
+
+
+            Shebang is always the first line in a script file (if it's there at all), and we can also check for it to check the interpreter
+            example:
+
+            #!/bin/python or #!/bin/perl or #!/bin/sh or #!/bin/python3
+            # tags = [core_approved, http, robotstxt, hyper, get_request]
```
bee-san, line 77: Ah you have  this structure here too, nice!

src/scripts/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+                todo!();
+            }
+        }
+    }
+
+    fn execute_script(self) -> Result<String> {
+        let process = Exec::cmd(self.command).args(&self.arguments);
+        match process.capture() {
+            Ok(c) => {
+                let es = match c.exit_status {
```
bee-san, line 122: Python scripts, by default, do not exit with OS status codes by the way -- we'd have to enforce this with the users submissions :)

src/scripts/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+                    return Err(anyhow!("exit code = {}", es));
+                }
+                Ok(c.stdout_str())
+            }
+            Err(error) => Err(anyhow!(error.to_string())),
+        }
+    }
+}
+
+struct ScriptConfig {
```
bee-san, line 138: I was thinking of moving the config and scripts to `~/.rustscan/` so:

```
~/.rustscan/config.toml
~/.rustscan/scripts/nmap.py
```

bee-san (commented):
Ideally, scripts should also be able to access the config file. I think that'd open up a significant amount of customisation.

Here's my idea:

```
/// Finds the directory of where the scripts are stored
/// and returns a vector of fully qualified paths to each script
fn find_scripts{

}

/// Given a vector of scripts, store the tags of each script
/// in a pre-defined struct so we can easily search / sort them later
/// Some tags include:
/// * Language
/// * Developer
/// * Core Approved
/// * How to call the script, so `python3 {locatinon} --ip {ip} --port {port}
/// Our script will fill in the {} for them
fn get_tags{

}

/// Given structs and the tag system
/// Calculate what the tag does
/// I.E the tag(s) might be "(Popular and Core) not Port 80"
/// This returns a vector of paths (or the structs) that meet the tag requirement
/// We will also need to check compatiability with script.
/// If script is Python, do we have Python on the system?
/// If script requires port 80, do we have port 80 open?
/// If the script needs a host name instead of IP, do we have that?
/// You get the idea -- we calculate them instead of the script because
/// multiple scripts may want Python, so we will only need to calculate it once
/// also reduces complexity for script creators
fn calculate_tags_get_scripts{

}

/// Runs the scripts according to how the user says to run them
fn run_scripts{

}
```

src/scripts/mod.rs
bergabman: Good idea, we can probably use absolute path, but if not we can change the working directory of RustScan on the fly and execute from the right directory with the right version of the interpreter.

src/scripts/mod.rs
bergabman: Parsed from the file itself, so it stay self-contained. We will have more possibilities to identify the interpreter, so it will be easy to fill that field.

src/scripts/mod.rs
bergabman: I think as long as it completes without error, it will return a `0`, every program ends with an exit systemcall, even if it is the interpreter. We should make some tests to see how it reacts to success and errors and what gets returned on the end. I will do that.

src/main.rs
bergabman: No this is just for maybe the ones we want to have built in. Like a default nmap scan. The rest of the scripts will be dynamic.. Maybe we don't need anything like that at all.

src/scripts/mod.rs
bergabman: We could do that, or just use a struct with predefined `Option`al fields, and parse from toml to that struct.

bee-san:
Can you resolve comments after they're done? and re-request code reviews when you need them too :D

src/scripts/mod.rs
bergabman: The ScriptConfig function now accepts a base path.

Cargo.toml
```diff
@@ -35,6 +35,10 @@ serde = "1.0.116"
 serde_derive = "1.0.116"
 cidr-utils = "0.5.0"
 itertools = "0.9.0"
+which = "4.0.2"
+anyhow = "1.0.32"
+subprocess = "0.2.6"
+text_placeholder = { version = "0.3", features = ["struct_context"] }
```
bee-san, line 41: Kaeros will love this :)

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    debug!("Scripts paths \n{:?}", script_paths);
+    let parsed_scripts = parse_scripts(script_paths);
+    debug!("Scripts parsed \n{:?}", parsed_scripts);
+
+    let script_config = ScriptConfig::read_config().unwrap();
+    debug!("Script config \n{:?}", script_config);
+
+    let mut scripts_to_run: Vec<ScriptFile> = Vec::new();
+
+    // Only Scripts that contain all the tags found in ScriptConfig will be selected.
```
bee-san, line 28: By tag do you mean stuff like "core" etc? or do you mean stuff like the author field too? If it's strictly tags like "core, http" etc i can see this being a neat lil security feature (an intended bug)

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+
+pub fn parse_scripts(scripts: Vec<PathBuf>) -> Vec<ScriptFile> {
+    let mut parsed_scripts: Vec<ScriptFile> = Vec::with_capacity(scripts.len());
+    for script in scripts {
+        let real_path = script.clone();
+        let mut lines_buf = String::new();
+        if let Ok(file) = File::open(script) {
+            for line in io::BufReader::new(file).lines().skip(1) {
+                if let Ok(mut line) = line {
+                    if line.starts_with("#") && line.len() > 3 {
```
bee-san, line 222: I take it we are only doing Python / Shell scripts, assuming that the lines start with `#`?
It's good for an MVP, but we need to be aware of this bug in the future :)

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+}
+
+/*
+Example ScriptFile headers, all fields are optional but we have to do the checks if something is necessary, for example an executor path.
+
+#!/usr/bin/perl
+#tags = ["core_approved", "bergabman"]
+#developer = [ "testdev", "https://example.org", "https://twitter" ]
+#execute_with = "python3"
+#join_ports = ","
```
bee-san, line 255: what does `","` mean here?

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+    parsed_scripts
+}
+
+/*
+Example ScriptFile headers, all fields are optional but we have to do the checks if something is necessary, for example an executor path.
+
+#!/usr/bin/perl
+#tags = ["core_approved", "bergabman"]
+#developer = [ "testdev", "https://example.org", "https://twitter" ]
+#execute_with = "python3"
```
bee-san, line 254: Why is this `#!/usr/bin/perl` but exex with python3? the script is in perl but the actual thing it's calling is python3?

bee-san (approved):

src/scripts/mod.rs
bergabman: The logic only collect those ScriptFiles from the scripts folder which contains all the tags predefined in the script_config file. For example you can give a crazy unique tag for a script and call that in the ScriptConfig file so only that script will run. We can build more logic around script parsing.

src/scripts/mod.rs
bergabman: Python, perl, bash all uses `#` as comment, but we for sure can use an extra variable in the ScriptConfig to define that character. That way it will be completely dyunamic, the users could add own characters.

src/scripts/mod.rs
bergabman: If that field is there, the found ports will be joined with that character and passed like that to the script. Like nmap uses `-p 80,8080,9001`

src/scripts/mod.rs
bergabman: My mistake :D

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+use anyhow::{anyhow, Result};
+use serde_derive::{Deserialize, Serialize};
+use std::fs::{self, File};
+use std::io::{self, prelude::*};
+use std::net::IpAddr;
+use std::path::PathBuf;
+use subprocess::{Exec, ExitStatus};
+use text_placeholder::Template;
+
+pub fn init_scripts() -> Vec<ScriptFile> {
```
bernardoamc, line 13: There is a common pattern in Rust and that is to initiate a struct resource through a method called `new`. So for this scenario we would obtain the scripts like we are already doing and inside `parse_scripts` we would call `ScriptFile::new()` that would contain the logic to parse this script. Let me know if this makes sense to you. 

```rust

```

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+            parsed_scripts.push(parsed);
+        } else {
+            debug!("Failed to parse toml");
+        }
+    }
+    parsed_scripts
+}
+
+/*
+Example ScriptFile headers, all fields are optional but we have to do the checks if something is necessary, for example an executor path.
```
bernardoamc, line 249: Do we need to move this to `ScriptFile` instead of `ScriptConfig` and document `ScriptConfig` itself?

src/input.rs
```diff
@@ -105,17 +105,21 @@ pub struct Opts {
     #[structopt(long, possible_values = &ScanOrder::variants(), case_insensitive = true, default_value = "serial")]
     pub scan_order: ScanOrder,
 
+    /// Use the top 1000 ports.
+    #[structopt(long)]
+    pub top: bool,
```
bernardoamc, line 110: Is this part of this PR?

src/main.rs
```diff
@@ -128,27 +133,45 @@ fn main() {
...
+                format!(
+                    "Script to be run {:?}, execution path {:?}, call format {:?}\n",
+                    script_f.execute_with, script_f.execute_with_path, script_f.call_format,
+                ),
+                opts.greppable.clone(),
+                opts.accessible.clone()
+            );
+
+            // This part allows us to add commandline arguments to the Script call_format, appending them to the end of the command.
+            if &opts.command.len() > &0 {
```
bernardoamc, line 148: Can we do `opts.command.len() > 0` here or this consumes the `command`?

src/main.rs
```diff
@@ -128,27 +133,45 @@ fn main() {
...
+                let user_nmap_args: Vec<String> = shell_words::split(&opts.command.join(" "))
```
bernardoamc, line 149: This is not necessarily `nmap` anymore, is it?

src/main.rs
```diff
@@ -128,27 +133,45 @@ fn main() {
         }
         detail!("Starting Nmap", opts.greppable, opts.accessible);
```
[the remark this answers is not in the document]
bergabman: I changed it to `Starting Script(s)`, that fits better, but we can also remove it of course.

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+        let mut files_vec: Vec<PathBuf> = Vec::new();
+        for entry in fs::read_dir(path)? {
+            let entry = entry?;
+            files_vec.push(entry.path());
+        }
+        return Ok(files_vec);
+    } else {
+        return Err(anyhow!("cant find scripts folder"));
+        // println!("Do you want me to create the {} directory?", &scripts_dir.to_string_lossy());
+        // std::fs::create_dir(&scripts_dir).unwrap();
```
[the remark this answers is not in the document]
bergabman: They are gone

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+        Err(error) => {
+            debug!("Command error {}", error.to_string());
+            return Err(anyhow!(error.to_string()));
+        }
+    }
+}
+
+pub fn find_scripts(mut path: PathBuf) -> Result<Vec<PathBuf>> {
+    path.push(".rustscan_scripts");
+    // check if dir is available
```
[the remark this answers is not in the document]
bergabman: Gone

src/input.rs
bergabman: This was part of an earlier PR, I just changed up the location of it.

bergabman:
After some discussions on discord, we are getting close to the final idea, (or maybe already there). Please check out the new implementation with examples and let me know what you think. @bee-san @bernardoamc 

It lacks some testing, hopefully we can make a Hacktoberfest issue from them.

src/scripts/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+            ip: ip,
+            ports: ports,
+            port: port,
+            join_ports: join_ports,
+            tags: tags,
+            call_format: call_format,
+        }
+    }
+
+    pub fn run(self) -> Result<String> {
```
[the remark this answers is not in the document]
bergabman: With the current structure I do consume part of the script so it's necessary as it is now.

src/scripts/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+#[derive(Clone, Debug)]
+pub struct Script {
+    // Path to the script itself.
+    path: Option<PathBuf>,
+
+    // Ip got from scanner.
+    ip: IpAddr,
+
+    // Ports found with portscan.
+    ports: Vec<u16>,
```
bernardoamc, line 116: What do you think of calling this `open_ports` since we now know they are open and calling the `port` option something like: `conditional_port` or `port_trigger`? Something that indicates that this port is the condition for the script to run.

src/scripts/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+    port: Option<String>,
+
+    // Character to join ports in case we want to use a string format of them, for example nmap -p.
+    join_ports: Option<String>,
+
+    // Tags found in ScriptFile.
+    tags: Option<Vec<String>>,
+
+    // The format how we want the script to run.
+    call_format: Option<String>,
```
bernardoamc, line 128: Shouldn't a script always have a `call_format`?

src/scripts/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+    port: String,
+}
+
+impl Script {
+    pub fn build(
+        path: Option<PathBuf>,
+        ip: IpAddr,
+        ports: Vec<u16>,
+        port: Option<String>,
+        join_ports: Option<String>,
```
bernardoamc, line 150: Should we call this `ports_separator`?

src/input.rs
```diff
@@ -199,6 +201,7 @@ pub struct Config {
     ulimit: Option<rlimit::rlim>,
     scan_order: Option<ScanOrder>,
     command: Option<Vec<String>>,
+    scripts: Option<String>,
```
bernardoamc, line 204: Since we know what to expect for these scripts we could use an `enum` here with the three options we expect. What do you think?

bernardoamc: This has the advantage that Rust will be able to enforce that we are always taking into account he three possible variants in our code.

src/scripts/mod.rs
bergabman: I changed the ports to open_ports so it's more descriptive, but I'm not sure in the conditional_port or port_trigger. What would you say for predefined_port? 🤔

src/scripts/mod.rs
bergabman: I was thinking on some error handling logic after parsing the script, but you are right, it shouldn't be an `Option`, and we can rely on the toml parser to throw an error. Now we just don't run the scripts simply that's not conform in any way. But we can stop it at parsing.

src/scripts/mod.rs
bergabman: `port_separator` is probably a better name, sounds ok, I'll change it.

src/input.rs
bergabman: All right, we can build an enum around the 3 values, now I just handle it like [ &_ => Ok(scripts_to_run),](https://github.com/RustScan/RustScan/pull/242/files#diff-5062b06fea1a548c03f86fb29d0fa7d9R103) :)

bergabman: Also, the commandline parser allows only the 3 options 
`#[structopt(long, possible_values = &["none", "default", "custom"], case_insensitive = true, default_value = "default")]`

bergabman:
I made the adjustments we discussed previously, changed `ports` we get from the scanner to `open_ports`, changed `join_ports` in structs and the right places in the example files to `ports_separator`. Created an Enum to parse `--scripta` user options from commandline and configfile with only those 3 values available, the same way `ScanOrder` is implemented. It was an easy copy thanks to the maker of that enum. Expanded the explanation part regarding the `execute_format` section.

I know somebody was talking about the placement of the script_config file, and also renaming it. But I can't seem to find that comment.

src/scripts/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+                        let script_hashset: HashSet<String> =
+                            script.tags.clone().unwrap().into_iter().collect();
+                        if config_hashset.is_subset(&script_hashset) {
+                            scripts_to_run.push(script.to_owned());
+                        }
+                        debug!(
+                            "\nScript tags does not match config tags {:?} {}",
+                            &script_hashset,
+                            script.path.clone().unwrap().display()
+                        );
```
bernardoamc, line 108: I think this is always running despite if there's a match or not.

src/scripts/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        let mut ports_str = String::new();
+        if let Some(ports_separator) = self.ports_separator {
+            let ports_str_vec: Vec<String> =
+                self.open_ports .iter().map(|port| port.to_string()).collect();
+            ports_str = ports_str_vec.join(&ports_separator);
+        } else if let None = self.ports_separator {
+            let ports_str_vec: Vec<String> =
+                self.open_ports .iter().map(|port| port.to_string()).collect();
+            ports_str = ports_str_vec.join(",");
```
bernardoamc, line 189: Here we can avoid the `if/else if` by doing:

```rust
let separator = self.ports_separator.unwrap_or(",");
let ports_str = self.open_ports .iter()
  .map(|port| port.to_string())
  .collect()
  .join(&ports_separator);
```

src/scripts/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+        } else if let Some(port) = self.port {
+            ports_str = port;
+        } else {
+            return Err(anyhow!("Failed to parse ports_separator in scriptfile."));
+        }
```
bernardoamc, line 194: Do we ever reach those since we are already taking into account the `Some` and `None` cases of `self.ports_separator`?

src/scripts/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+            let exec_parts: ExecParts = ExecParts {
+                ip: self.ip.to_string(),
+                port: ports_str,
+            };
+            to_run = default_template.fill_with_struct(&exec_parts)?;
+        }
+
+        debug!("\nTo run {}", to_run);
+
+        let arguments: Vec<String> = to_run.split(" ").map(|arg| arg.to_string()).collect();
```
bernardoamc, line 222: We should probably use `shell_words::split` here instead of a normal split by whitespace to avoid edge cases of argument parsing.

Cargo.toml
```diff
@@ -36,6 +36,9 @@ serde_derive = "1.0.116"
 cidr-utils = "0.5.0"
 itertools = "0.9.0"
 trust-dns-resolver = { version = "0.19.5", features = ["dns-over-rustls"] }
+anyhow = "1.0.32"
+subprocess = "0.2.6"
+text_placeholder = { version = "0.3", features = ["struct_context"] }
```
bernardoamc, line 41: Now we have a version `0.4` that uses less memory 😄

fixtures/test_rustscan_scripts.toml
```diff
@@ -0,0 +1,14 @@
+# Test/Example ScriptConfig file
+
+# Tags to filter on scripts. Only scripts containing all these tags will run.
+tags = ["core_approved", "example"]
```
bernardoamc, line 4: Are we over-complicating this tags system? Maybe we can ignore `tags` and `developer` completely until we actually allow people to contribute scripts. This will keep our code simpler and allow us to evolve it when we have the need to.

fixtures/test_script.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/python3
+#tags = ["core_approved", "example",]
+#developer = [ "example", "https://example.org" ]
+#port = "80"
+#call_format = "python3 {{script}} {{ip}} {{port}}"
```
bernardoamc, line 5: Now that I'm thinking about it we should probably make those independent of the script itself since we are preventing other scripts that do not support `#` as a comment from running.

src/scripts/mod.rs
bergabman: Yes, I forget to wrap it back to the else statement. Good catch, fixed it.

src/scripts/mod.rs
bergabman: It needed some type at `collect::<Vec<String>>()`, but it looks much better. Thanks!

Cargo.toml
bergabman: Nice! I bumped it. :)

bee-san:
@bergabman conflicts! :( But other than conflicts, how's this coming? Are we ready to release? Do you need any help on anything? :heart:

fixtures/test_script.py
bergabman: For the first release, I think we are good with accepting the files which do accept that as a comment. We can of course expand the parsing of ScriptFiles to include other marks.

src/scripts/mod.rs
bergabman: I changed to use the original `shell_words::split` parser for the commandline args.

src/scripts/mod.rs
bergabman: The separator got simplified to `let separator = self.ports_separator.unwrap_or(",".into());` based on your other comment. So now we can get here.

src/scripts/mod.rs
bergabman: I changed the variable name in the script files and the struct to `trigger_port`, I think that describes it good.

src/scripts/mod.rs
bergabman: I implemented a `ScriptFile::new()` function, and moved the parsing logic to there. So the ScriptFile gets parsed and return an `Option`, and gets filtered on `Some()` and `None`. We can expand the information we give to the user in the future to be more clear if sommething goes wrong. Now it just gets discarded.

bernardoamc:
There are a few details/performance improvements that we can tackle, but if we are happy with the functionality we can always merge this and keep iterating on it. Great job on this @bergabman <3 

The things that we can tackle later on:

1. Reduce memory allocation through usage of references and lifetimes
    * Strings
    * Hashset
2. Keep improving the scripts usage through things like port matching and so on

src/main.rs
```diff
@@ -112,46 +118,71 @@ fn main() {
...
+                script_f.ports_separator,
+                script_f.tags,
+                script_f.call_format,
+            );
+            match script.run() {
+                Ok(script_result) => {
+                    detail!(
+                        format!("{}", script_result),
+                        opts.greppable.clone(),
+                        opts.accessible.clone()
```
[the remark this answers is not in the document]
bergabman: No we don't, for some reason I ended up cloning them during development. I think the struct itself didn't have `Copy` implemented at the beginning. Fixed both.

src/main.rs
```diff
@@ -165,7 +196,7 @@ fn print_opening(opts: &Opts) {
 | {}  }| { } |{ {__ {_   _}{ {__  /  ___} / {} \ |  `| |
 | .-. \| {_} |.-._} } | |  .-._} }\     }/  /\  \| |\  |
 `-' `-'`-----'`----'  `-'  `----'  `---' `-'  `-'`-' `-'
-Faster Nmap scanning with Rust."#;
+The Modern Day Port Scanner."#;
```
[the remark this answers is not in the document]
bergabman: I know right :)

src/main.rs
```diff
@@ -112,46 +118,71 @@ fn main() {
...
+                        format!("{}", script_result),
+                        opts.greppable.clone(),
+                        opts.accessible.clone()
+                    );
+                }
+                Err(e) => {
+                    warning!(
+                        &format!("Error {}", e.to_string()),
+                        opts.greppable.clone(),
+                        opts.accessible.clone()
```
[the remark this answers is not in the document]
bergabman: You are right, see other comment.

src/main.rs
```diff
@@ -112,46 +118,71 @@ fn main() {
...
+                opts.accessible.clone()
+            );
 
-        // Runs the nmap command and spawns it as a process.
-        let mut child = Command::new("nmap")
-            .args(&nmap_args)
-            .spawn()
-            .expect("failed to execute nmap process");
+            // This part allows us to add commandline arguments to the Script call_format, appending them to the end of the command.
+            if opts.command.len() > 0 {
```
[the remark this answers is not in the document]
bergabman: I think it makes sense, in case the user want to simply append an extra argument to the execution of the scripts. For example in case of an nmap script, we give the possibility to just append extra arguments to RustScan that gets passed to the script. An extra `-sC` switch for nmap will be just a quick append to the RustScan commandline arguments. If we remove it, the users will have to change the script file every time they want to append a new argument to it. This will also improve development time and experience for writing RustScan script files, as the writer of the script can test arguments in terminal and append them to the script execution format if they fit.

bergabman:
I improved the PR based on the reviews. I think this version of the ScriptingEngine will be a good start to let people write custom scripts for RustScan. It certainly can be improved and as @bernardoamc said we should tackel efficiency points like using references and lifetimes. Luckily we don't have to parse a lot of script files as of now, and I can imagine people will most likely run 1 script with the results of the portscan, so memory allocations are not that big of a deal. But to have a more elegant and correct code we will have to tackle those points. 

The module lacks testing, but the example files in the `fixtures` directory can be used to write the tests. There is an example script config file and 4 different scripts to work with. Maybe it's worth to create an issue to ask people to write tests for us :).

bee-san:
Considering this issue is for an initial RSE, I see no reason why this shouldn't be merged :) The most important thing now should be writing docs on how to use the thing :sweat_smile:

Issue #245, opened 2020-09-27
DNS Resolver does not use OS' resolver


bee-san:
**Describe the bug**
Per `trust_dns_resolver`'s documentation:
https://docs.rs/trust-dns-resolver/0.19.5/trust_dns_resolver/

> It does not use the Host OS' resolver. If what is desired is to use the Host OS' resolver, generally in the system's libc, then the std::net::ToSocketAddrs variant over &str should be used.

This causes some problems as with CTFs sometimes we have to change our hosts file to reach the target webpage. 

Also, it would be faster to use the OS' resolver than an external one.

bee-san:
Opening this up to Hacktoberfest <3 Please leave a comment saying you want this and I'll assign you :D

shahar481:
Hey :),
So basically you just want to replace all resolving you did with trust dns with to_socket_addrs?
I'll start working on it.

bee-san:
> Hey :),
> So basically you just want to replace all resolving you did with trust dns with to_socket_addrs?
> I'll start working on it.

Hey! Sorry for not replying, I was asleep.

I didn't explain this issue very well, my bad. 

We want to keep both trust_dns but use `std::net::ToSocketAddrs` at the same time, I think.

The logic should be:

* Use `std::net::ToSocketAddrs` to check the host OS' resolver.
* If it didn't resolve:
    * Use Trust DNS

So we use the local Host Resolver on the PC (the hosts file, I believe it's called) but if it's not in there we use Trust DNS -- do you understand that? Or am I being confusing again. My bad! I'm quite tired right now :sweat_smile:

shahar481:
Yeah I think I got it. You want trust-dns to be the fallback resolver if the OS couldn't resolve.

shahar481:
Hey :D, 
Don't want to ping you too much. Just letting you know that I updated the PR.

Issue #238, opened 2020-09-22
Host file assumes host is IP address only


bee-san:
Because of the way the host file works, it can only work on IP addresses.

```rust
    for ip_or_host in &opts.addresses {
        match IpCidr::from_str(ip_or_host) {
            Ok(cidr) => cidr.iter().for_each(|ip| ips.push(ip)),
            _ => match format!("{}:{}", &ip_or_host, 80).to_socket_addrs() {
                Ok(mut iter) => ips.push(iter.nth(0).unwrap().ip()),
                _ => {
                    match read_ips_from_file(ip_or_host.to_owned()) {
                        Ok(x) => ips.extend(x),
                        _ => {
                            warning!(
                                format!("Host {:?} could not be resolved.", ip_or_host),
                                opts.greppable,
                                opts.accessible
                            );
                        },
                    }

                }
            },
        }

    }
```

So the code checks:
1. CIDR
2. Host
3. File
In this order.

But if the file contains:

```
google.com
192.168.0.0/30
```

This will not work, as those conversions are done before the file.

ta1bbty:
#445

PR #246, opened 2020-09-27
Added DNS lookup using Google DNS, refactored IP code

### LINKED ISSUE #239: Use DNS instead of socket connection
Currently RustScan connects to port 80 to resolve the host using sockets, DNS would be much nicer to use.

bee-san:
* Added DNS lookup. We're using Google for DNS. This brings 2 problems:
1. Google bad
2. No DNS over HTTPS with Google DNS
However, the advantage is:
1. It's noticeably faster than Cloudflare.

* Refactored IP parsing code

Now we have a difference between what could be an IP, and what isn't. For instance, the code now does this:
* Checks for CIDR
* Checks for IP:port
* Checks for Host (DNS resolution)
This returns a vector of IPs (CIDR / DNS return vectors).

Our file input runs this function against every possible IP, generating a vector. After each run of `parse_single_ip_or_host`, it merges the two vectors together.

This fixes #239 #238 but introduces bug #245


src/main.rs
```diff
@@ -208,6 +209,47 @@ fn build_nmap_arguments<'a>(
...
+        }
+    }
+    println!("Ips is {:?}", ips);
+    ips
+}
+
+/// Uses DNS to get the IPS assiocated with host
+fn get_ips_from_host(ip_or_host: &String) -> Result<Vec<IpAddr>, std::io::Error> {
+    let mut ips: Vec<std::net::IpAddr> = Vec::new();
+    let resolver = Resolver::new(ResolverConfig::default(), ResolverOpts::default()).unwrap();
```
bee-san, line 240: I dislike how every time this function is called, I create a new resolver. I couldn't get it to work via arguments (it doesn't implement a lot of traits and `cargo check` hated it.

The alternative is a global variable, unless someone can make a further PR fixing this :)

src/main.rs
```diff
@@ -208,6 +209,47 @@ fn build_nmap_arguments<'a>(
     arguments
 }
 
+/// Goes through all possible IP inputs (files or via argparsing)
+/// Parses the string(s) into IPs
+fn parse_addresses(opts: &Opts) -> Vec<IpAddr> {
+    // Construct a new Resolver with default configuration options, DNS.
```
bee-san, line 215: Dead comment, need to delete this. Will delete after the next comment is talked about :)

src/main.rs
bee-san: Never mind! This might have been a different bug and I confused the two. Seems to be working just fine now :)

src/main.rs
```diff
@@ -208,6 +209,47 @@ fn build_nmap_arguments<'a>(
...
+    let mut ips: Vec<IpAddr> = Vec::new();
+
+    for ip_or_host in &opts.addresses {
```
bernardoamc, line 217: What do you think of renaming `ip_or_host` to `source`?

src/main.rs
```diff
@@ -208,6 +209,47 @@ fn build_nmap_arguments<'a>(
...
+                }
+            },
+        }
+    }
+    println!("Ips is {:?}", ips);
+    ips
+}
+
+/// Uses DNS to get the IPS assiocated with host
+fn get_ips_from_host(ip_or_host: &String) -> Result<Vec<IpAddr>, std::io::Error> {
```
bernardoamc: What do you think of naming this `resolve_ips_from_host`?

src/main.rs
```diff
@@ -218,37 +260,29 @@ fn read_ips_from_file(ips: String) -> Result<Vec<std::net::IpAddr>, std::io::Err
     let mut ips: Vec<std::net::IpAddr> = Vec::new();
 
     for str_ip in reader.lines() {
-        match IpAddr::from_str(&str_ip.unwrap()) {
-            Ok(x) => ips.push(x),
-            Err(y) => panic!("File does not contain valid IP address. Error at {}", y),
-        }
+        // This code assumes that every line in the file is a valid
```
bernardoamc, line 264: Do we want to make this assumption or instead log the invalid IP and skip it?

src/main.rs
```diff
@@ -218,37 +260,29 @@ fn read_ips_from_file(ips: String) -> Result<Vec<std::net::IpAddr>, std::io::Err
...
+        ips.extend(parse_single_ip_or_host(str_ip.unwrap()).unwrap());
     }
     Ok(ips)
 }
 
-fn parse_addresses(opts: &Opts) -> Vec<IpAddr> {
+/// Given a string, parse it as an host, IP address, or CIDR.
+/// This allows us to pass files as hosts or cidr or IPs easily
+/// Call this everytime you have a possible IP_or_host
+fn parse_single_ip_or_host(ip_or_host: String) -> Result<Vec<IpAddr>, std::io::Error> {
```
bernardoamc, line 273: What do you think of calling this `parse_to_ip` and rename the argument to `address`?

bee-san:
ToDo still:
- [ ] Add tests for fixtures

src/main.rs
```diff
@@ -419,4 +467,10 @@ mod tests {
 
         assert_eq!(ips.is_empty(), true);
     }
+    #[test]
+    fn parse_hosts_file_and_incorrect_hosts() {
```
bernardoamc, line 471: We are missing an assertion here ( `assert_eq` ) with the ips that were parsed correctly. 😄

bee-san: Good catch! Fixed.

bee-san: Just for future reference, the current way of testing domains might fail.

Our DNS resolver can return multiple IPs for 1 domain name, and we cannot guarantee that Google or example.com will return 1 IP each time.

bernardoamc: Good point 😄 I will revisit the way we test at some point, ideally our tests shouldn't fail when an external party fails.

bernardoamc (approved):

SuperSandro2000:
> It's noticeably faster than Cloudflare.

This is just a hoax. It depends on location and for me google is 10 times slower than cloudflare.

bee-san:
> > It's noticeably faster than Cloudflare.
> 
> This is just a hoax. It depends on location and for me google is 10 times slower than cloudflare.

You're right! I'll run some benchmark tests. :)

PR #249, opened 2020-09-29
Docker: Fix cosmetic issues, change maintainer to rustscan

SuperSandro2000:
Changing only a few lines does not give you maintainer. Also people might contact the person if they need to reach out about something which is probably not intended.


CMNatic:
Heya @SuperSandro2000 I agree for the most part in the sense that no, changing a few lines doesn't make you a maintainer, in the same way, it doesn't make you the new author.

I added that label as at the moment I'm the best point of contact re anything Docker for the core team where I'm responsible for bringing RustScan to Docker, maintaining it and improving the CI process, so the label was a reflection of that. 

It's especially important that we ensure that credit was kept to Hydragum for the core of the current Dockerfile which serves as a fantastic starting point for us whilst I rework the `Dockerfile` and Debian-builder `Dockerfile`.

Hindsight, it's probably best to change the maintainer label to RustScan as you suggest so that issues can be raised in the right repository (aka here). Adding a `vendor=rustscan` label wouldn't hurt either perhaps?

##  Re: Changing `COPY` Contexts
I'm opposed to this change in the current state of the `Dockerfile` and here was my logic about it:

1) The context at this stage is already in `/usr/src/rustscan` via `WORKDIR` and so we would only be using relative paths as the destination, at this stage, `src` does not exist as a directory in the container.

2) `COPY` is capable of detecting the source file as a directory from the host - where the contents are copied recursively to the destination in the container, this is why the source `src` has been left without the trailing slash.

> If <src> is a directory, the entire contents of the directory are copied, including filesystem metadata. (https://docs.docker.com/engine/reference/builder/#copy)

3) The destination `src` has been left without the trailing `/`  (despite it being practice to have it when using commands like `cp` and `mv` on the terminal) because otherwise, this would copy `src` into `/usr/src/rustscan/src/src` (unless we make the `/usr/src/rustscan/src`) directory first before copying from my understanding of how the various `COPY` works.

> If <dest> does not end with a trailing slash, it will be considered a regular file and the contents of <src> will be written at <dest>. (https://docs.docker.com/engine/reference/builder/#copy)

Remembering that our src is a directory that and as such would get recursively copied. If we were copying individual files then we'd need to use a trailing `/` on the destination `src`

## Re: EOL Warning
This is a whoopsie by using Git on Windows, can be solved amongst another commit (: good spot however

What are your thoughts on this? Thanks! 👍🏻

SuperSandro2000:
> Adding a vendor=rustscan label wouldn't hurt either perhaps?

I don't really know.

> Docker for the core team where I'm responsible for bringing RustScan to Docker

None of the RustScan org members are public which does not make it very easy to find them.

> This is a whoopsie by using Git on Windows

I just configured my VSCode/Vim to always append a final new line and trim to many. Works great when working with git.

> The destination src has been left without the trailing / (despite it being practice to have it when using commands like cp and mv on the terminal) because otherwise, this would copy src from the host into /usr/src/rustscan/src/src (unless we make the /usr/src/rustscan/src) directory first before copying from my understanding of how the various COPY works.

I just tested it locally and it works fine. Docker maybe detects if the directory names are the same and does not nest them. Not sure on the details but it works.

I like to add the final ``/`` to make sure the things copied are directories and not files. I think I once debugged such issue for an hour and since than I am doing it. Probably not to important but I think it is nice to have.

CMNatic:
Sounds good to me, both ways work just as well atm (: . Your change by manually specifying it rather than relying on Docker to work as expected is good for future debugging. Thanks for discussing your PR and taking the time to contribute to RustScan 💯

PR #259, opened 2020-10-03
Added more boundary tests for scanners mod.rs

bee-san:
Added more tests to scanner, in particular:
- Backwards ranges (10 - 1)
- No range (0 - 0)
- High range (1 - 65535)
- Serial testing (tests were random only before)


src/scanner/mod.rs
```diff
@@ -232,4 +232,49 @@ mod tests {
...
+    }
+    #[test]
+    fn scanner_doesnt_run_no_range() {
+        let addrs = vec!["8.8.8.8".parse::<IpAddr>().unwrap()];
+
+        let range = PortRange { start: 0, end: 0 };
+        let strategy = PortStrategy::pick(Some(range), None, ScanOrder::Serial);
+        let scanner = Scanner::new(&addrs, 10, Duration::from_millis(100), true, strategy, true);
+        block_on(scanner.run());
+        assert_eq!(1, 1);
```
bernardoamc, line 256: What do you think of asserting that we don't have open ports for the cases where we don't expect the scanner to run?

bernardoamc (approved):
Left a small comment as an idea, thanks for the extra tests bee. :)

bernardoamc:
We seem to have file conflicts, hope they are not too hard to fix 😄

bee-san:
> We seem to have file conflicts, hope they are not too hard to fix smile

Ah, they are because I git reset one too many times :sweat_smile:

PR #263, opened 2020-10-05
Fix archives-generation from Makefile 👷‍♂️

TGotwig:
removes all that nesting inside the archives


bee-san:
Hey @TGotwig ! I remember you, the only one to add emojis to their commits ;)

Thanks so much again for this commit! :D Your support is much appreciated :rocket:

bee-san (approved):

PR #264, opened 2020-10-06
Changed the dns resolving to use std::net::ToSocketAddrs

### LINKED ISSUE #245: DNS Resolver does not use OS' resolver

shahar481:
* Made trust-dns a fallback resolver if the OS can't resolve

Fixes #245


src/main.rs
```diff
@@ -283,14 +290,14 @@ fn read_ips_from_file(
 /// Given a string, parse it as an host, IP address, or CIDR.
 /// This allows us to pass files as hosts or cidr or IPs easily
 /// Call this everytime you have a possible IP_or_host
-fn parse_to_ip(address: String, resolver: &Resolver) -> Result<Vec<IpAddr>, std::io::Error> {
```
niklasmohrin, line 286: This functions signature should also be changed: `address` should be `&str`, because ownership is not needed. The return type should be just the vector as the function always returns `Ok`.

niklasmohrin (commented):
Hey there! Thanks for contributing!

Every logic you introduced is correct as I see it. I did however not check whether there are other places that need this addition too yet, but I don't think there are some anyway.

Most of the stuff I suggest are lints from `cargo clippy` or style changes (with performance gains sometimes). Interestingly, most of the things have been here already and are just things that need fixing in general. This PR is a great opportunity to fix these questionable style choices.

bee-san:
@all-contributors please add @shahar481 and @niklasmohrin for code

src/main.rs
```diff
@@ -213,13 +213,13 @@ fn build_nmap_arguments<'a>(
...
         &Resolver::new(ResolverConfig::cloudflare_tls(), ResolverOpts::default()).unwrap();
 
     for ip_or_host in &opts.addresses {
-        match read_ips_from_file(ip_or_host.to_owned(), &resolver) {
+        match read_ips_from_file(ip_or_host.to_owned(), &backup_resolver) {
             Ok(x) => ips.extend(x),
-            _ => match parse_to_ip(ip_or_host.to_owned(), &resolver) {
+            _ => match parse_to_ip(ip_or_host.to_owned(), &backup_resolver) {
                 Ok(x) => ips.extend(x),
                 _ => {
```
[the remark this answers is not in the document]
shahar481: Yeah you are absolutely right. But I think this PR is not the place for this change. I'll refactor the dns resolving part of this PR. Afterwards I'll make a new PR where I will refactor all the match expressions and improve the overall style of the code. 

Would love to hear your thoughts on this. If you disagree I can add all that to this PR.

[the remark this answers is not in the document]
niklasmohrin: It's up to you. I think it would be nice to not merge a PR with any "new" clippy warnings / clippy warnings that are on the diff, but you sure don't have to include these fixes here. The middle way would be to have one commit that introduces the DNS stuff and a second one that cleans up the diff, but it sure is your decision.

niklasmohrin:
@shahar481 Some PRs just went live, so you won't have to do most of these chores. Please still consider doing the rest of them 😄

shahar481:
@niklasmohrin I think I fixed all the problems. Feel free to point out more random things to fix if you see any :)

src/main.rs
```diff
@@ -274,18 +274,27 @@ fn parse_address(address: &str, resolver: &Resolver) -> Vec<IpAddr> {
...
-fn resolve_ips_from_host(source: &str, resolver: &Resolver) -> Vec<IpAddr> {
-    resolver
-        .lookup_ip(source)
-        .map(|x| x.iter().collect())
-        .unwrap_or_default()
+fn resolve_ips_from_host(source: &str, backup_resolver: &Resolver) -> Vec<IpAddr> {
+    let mut ips: Vec<std::net::IpAddr> = Vec::new();
+
+    if let Ok(addrs) = source.to_socket_addrs() {
+        for ip in addrs.into_iter() {
```
niklasmohrin, line 281:
```suggestion
        for ip in addrs {
```

src/main.rs
```diff
@@ -274,18 +274,27 @@ fn parse_address(address: &str, resolver: &Resolver) -> Vec<IpAddr> {
...
+            ips.push(ip.ip());
+        }
+    } else if let Ok(addrs) = backup_resolver.lookup_ip(&source) {
+        for ip in addrs.iter() {
+            ips.push(ip);
+        }
```
niklasmohrin, line 287: Here you can just use `.extend()`

niklasmohrin (commented):
You were right, it's just small things again :) The rest looks good :+1:

shahar481:
Added the changes from the CR :)

niklasmohrin:
@shahar481 Looks great! Very concise PR overall :+1:

bee-san (approved):

bee-san:
@all-contributors please add @shahar481 for code

PR #266, opened 2020-10-06
Add tries to scan_socket

### LINKED ISSUE #38: Increase max retries limit
Increase the amount of times RistScan retries scanning before giving up

eiffel-fl:
Hi.

I added a number of tries to `scan_socket`, by default the number of try is 1 but it can be set with command line option `--tries`.
This should close #38.

I am not a Rust expert, so please let me know if I wrote something wrong or not elegant.
Moreover, I am not sure of my code for the `Config` part, so this part needs to be reviewed.

Best regards.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -95,6 +95,11 @@ pub struct Opts {
     #[structopt(short, long, default_value = "1500")]
     pub timeout: u32,
 
+    /// The number of tries before a port is assumed to be closed.
+    /// If set to 0, rustscan will correct it to 1.
+    #[structopt(long, default_value = "1")]
+    pub tries: u8,
+
     /// Automatically ups the ULIMIT with the value you provided.
     #[structopt(short, long)]
     pub ulimit: Option<rlimit::rlim>,
@@ -153,7 +158,9 @@ impl Opts {
             }
         }
 
-        merge_required!(addresses, greppable, accessible, batch_size, timeout, scan_order, command);
+        merge_required!(
+            addresses, greppable, accessible, batch_size, timeout, tries, scan_order, command
+        );
     }
 
     fn merge_optional(&mut self, config: &Config) {
@@ -195,6 +202,7 @@ pub struct Config {
     accessible: Option<bool>,
     batch_size: Option<u16>,
     timeout: Option<u32>,
+    tries: Option<u8>,
     no_nmap: Option<bool>,
     ulimit: Option<rlimit::rlim>,
     scan_order: Option<ScanOrder>,
@@ -252,6 +260,7 @@ mod tests {
                 greppable: Some(true),
                 batch_size: Some(25_000),
                 timeout: Some(1_000),
+                tries: Some(1),
                 ulimit: None,
                 no_nmap: Some(false),
                 command: Some(vec!["-A".to_owned()]),
@@ -270,6 +279,7 @@ mod tests {
                 greppable: true,
                 batch_size: 0,
                 timeout: 0,
+                tries: 0,
                 ulimit: None,
                 command: vec![],
                 accessible: false,
--- a/src/scanner/mod.rs
+++ b/src/scanner/mod.rs
@@ -11,6 +11,7 @@ use futures::stream::FuturesUnordered;
 use std::{
     io::ErrorKind,
     net::{IpAddr, Shutdown, SocketAddr},
+    num::NonZeroU8,
     time::Duration,
 };
 
@@ -26,6 +27,7 @@ pub struct Scanner {
     ips: Vec<IpAddr>,
     batch_size: u16,
     timeout: Duration,
+    tries: NonZeroU8,
     greppable: bool,
     port_strategy: PortStrategy,
     accessible: bool,
@@ -36,6 +38,7 @@ impl Scanner {
         ips: &[IpAddr],
         batch_size: u16,
         timeout: Duration,
+        tries: u8,
         greppable: bool,
         port_strategy: PortStrategy,
         accessible: bool,
@@ -43,6 +46,7 @@ impl Scanner {
         Self {
             batch_size,
             timeout,
+            tries: NonZeroU8::new(std::cmp::max(tries, 1)).unwrap(),
             greppable,
             port_strategy,
             ips: ips.iter().map(|ip| ip.to_owned()).collect(),
@@ -87,54 +91,66 @@ impl Scanner {
         open_sockets
     }
 
-    /// Given a port, scan it.
+    /// Given a socket, scan it self.tries times.
     /// Turns the address into a SocketAddr
     /// Deals with the <result> type
     /// If it experiences error ErrorKind::Other then too many files are open and it Panics!
-    /// ese any other error, it returns the error in Result as a string
-    /// If no  errors occur, it returns the port number in Result to signify the port is open.
+    /// Else any other error, it returns the error in Result as a string
+    /// If no errors occur, it returns the port number in Result to signify the port is open.
     /// This function mainly deals with the logic of Results handling.
     /// # Example
     ///
-    ///     self.scan_port(10:u16)
+    ///     self.scan_socket(socket)
     ///
     /// Note: `self` must contain `self.ip`.
     async fn scan_socket(&self, socket: SocketAddr) -> io::Result<SocketAddr> {
-        match self.connect(socket).await {
-            Ok(x) => {
-                debug!(
-                    "Connection was successful, shutting down stream {}",
-                    &socket
-                );
-                match x.shutdown(Shutdown::Both) {
-                    Err(e) => debug!("Shutdown stream error {}", &e),
-                    _ => {}
-                }
-                if !self.greppable {
-                    if self.accessible {
-                        println!("Open {}", socket.to_string());
-                    } else {
-                        println!("Open {}", socket.to_string().purple());
-                    }
-                }
+        let tries = self.tries.get();
 
-                Ok(socket)
-            }
-            Err(e) => match e.kind() {
-                ErrorKind::Other => {
-                    if e.to_string().contains("No route to host")
-                        || e.to_string().contains("Network is unreachable")
-                    {
-                        debug!("Socket connect error: {} {}", &e.to_string(), &socket);
-                        Err(io::Error::new(io::ErrorKind::Other, e.to_string()))
-                    } else {
-                        debug!("Socket connect error: {} {}", &e.to_string(), &socket);
-                        panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
+        debug!("self.tries: {}", tries);
+
+        for nr_try in 1..=tries {
+            debug!("Try number: {}", nr_try);
+
+            match self.connect(socket).await {
+                Ok(x) => {
+                    debug!(
+                        "Connection was successful, shutting down stream {}",
+                        &socket
+                    );
+                    if let Err(e) = x.shutdown(Shutdown::Both) {
+                        debug!("Shutdown stream error {}", &e);
+                    }
+                    if !self.greppable {
+                        if self.accessible {
+                            println!("Open {}", socket.to_string());
+                        } else {
+                            println!("Open {}", socket.to_string().purple());
+                        }
+                    }
+
+                    debug!("Return Ok after {} tries", nr_try);
+                    return Ok(socket);
+                }
+                Err(e) => {
+                    let error_string = e.to_string();
+
+                    if e.kind() == ErrorKind::Other {
+                        debug!("Socket connect error: {} {}", &error_string, &socket);
+
+                        if !error_string.contains("No route to host")
+                            && !error_string.contains("Network is unreachable")
+                        {
+                            panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
+                        }
+                    }
+
+                    if nr_try == tries {
+                        return Err(io::Error::new(io::ErrorKind::Other, error_string));
                     }
                 }
-                _ => Err(io::Error::new(io::ErrorKind::Other, e.to_string())),
-            },
+            };
         }
+        unreachable!();
     }
 
     /// Performs the connection to the socket with timeout
@@ -174,7 +190,15 @@ mod tests {
             end: 1_000,
         };
         let strategy = PortStrategy::pick(Some(range), None, ScanOrder::Random);
-        let scanner = Scanner::new(&addrs, 10, Duration::from_millis(100), true, strategy, true);
+        let scanner = Scanner::new(
+            &addrs,
+            10,
+            Duration::from_millis(100),
+            1,
+            true,
+            strategy,
+            true,
+        );
         block_on(scanner.run());
         // if the scan fails, it wouldn't be able to assert_eq! as it panicked!
         assert_eq!(1, 1);
@@ -188,7 +212,15 @@ mod tests {
             end: 1_000,
         };
         let strategy = PortStrategy::pick(Some(range), None, ScanOrder::Random);
-        let scanner = Scanner::new(&addrs, 10, Duration::from_millis(100), true, strategy, true);
+        let scanner = Scanner::new(
+            &addrs,
+            10,
+            Duration::from_millis(100),
+            1,
+            true,
+            strategy,
+            true,
+        );
         block_on(scanner.run());
         // if the scan fails, it wouldn't be able to assert_eq! as it panicked!
         assert_eq!(1, 1);
@@ -201,7 +233,15 @@ mod tests {
             end: 1_000,
         };
         let strategy = PortStrategy::pick(Some(range), None, ScanOrder::Random);
-        let scanner = Scanner::new(&addrs, 10, Duration::from_millis(100), true, strategy, true);
+        let scanner = Scanner::new(
+            &addrs,
+            10,
+            Duration::from_millis(100),
+            1,
+            true,
+            strategy,
+            true,
+        );
         block_on(scanner.run());
         assert_eq!(1, 1);
     }
@@ -213,7 +253,15 @@ mod tests {
             end: 445,
         };
         let strategy = PortStrategy::pick(Some(range), None, ScanOrder::Random);
-        let scanner = Scanner::new(&addrs, 10, Duration::from_millis(100), true, strategy, true);
+        let scanner = Scanner::new(
+            &addrs,
+            10,
+            Duration::from_millis(100),
+            1,
+            true,
+            strategy,
+            true,
+        );
         block_on(scanner.run());
         assert_eq!(1, 1);
     }
@@ -228,7 +276,15 @@ mod tests {
             end: 600,
         };
         let strategy = PortStrategy::pick(Some(range), None, ScanOrder::Random);
-        let scanner = Scanner::new(&addrs, 10, Duration::from_millis(100), true, strategy, true);
+        let scanner = Scanner::new(
+            &addrs,
+            10,
+            Duration::from_millis(100),
+            1,
+            true,
+            strategy,
+            true,
+        );
         block_on(scanner.run());
         assert_eq!(1, 1);
     }
--- a/README.md
+++ b/README.md
@@ -242,6 +242,8 @@ OPTIONS:
                                      ascending order while the "random" option will scan ports randomly [default:
                                      serial]  [possible values: Serial, Random]
     -t, --timeout <timeout>          The timeout in milliseconds before a port is assumed to be closed [default: 1500]
+        --tries <tries>              The number of tries before a port is assumed to be closed. If set to 0, rustscan
+                                     will correct it to 1 [default: 1]
     -u, --ulimit <ulimit>            Automatically ups the ULIMIT with the value you provided
 
 ARGS:
@@ -260,7 +262,7 @@ Your operating system may not support this, but it is worth it to play around an
 
 ## Configuration file
 
-This binary accepts a configuration file that is read from the home directory of the user. It follows the TOML format
+This binary accepts a configuration file, named `.rustscan.toml`, that is read from the home directory of the user. It follows the TOML format
 and accepts the following fields:
 
 - `addresses`
@@ -272,6 +274,7 @@ and accepts the following fields:
 - `greppable`
 - `batch-size`
 - `timeout`
+- `tries`
 - `ulimit`
 
 ### Format example
@@ -286,6 +289,7 @@ accessible = true
 scan_order = "Serial"
 batch_size = 1000
 timeout = 1000
+tries = 3
 ulimit = 1000
 ```
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -77,6 +77,7 @@ fn main() {
         &ips,
         batch_size,
         Duration::from_millis(opts.timeout.into()),
+        opts.tries,
         opts.greppable,
         PortStrategy::pick(opts.range, opts.ports, opts.scan_order),
         opts.accessible,


src/scanner/mod.rs
```diff
@@ -100,41 +103,59 @@ impl Scanner {
...
-                match x.shutdown(Shutdown::Both) {
-                    Err(e) => debug!("Shutdown stream error {}", &e),
-                    _ => {}
-                }
-                if !self.greppable {
-                    if self.accessible {
-                        println!("Open {}", socket.to_string());
-                    } else {
-                        println!("Open {}", socket.to_string().purple());
+        let mut ret = Err(io::Error::new(
```
niklasmohrin, line 106: This variable is not needed if what I propose in the other comments is implemented. However, the default error still has to be returned if `self.tries` is 0, but you can move that behind the for loop, which will also avoid the allocation with `format!`, if it is not used.

src/scanner/mod.rs
```diff
@@ -100,41 +103,59 @@ impl Scanner {
...
+            io::ErrorKind::Other,
+            format!("Did not connect to {}", &socket),
+        ));
+
+        for nr_try in 0..self.tries {
```
niklasmohrin, line 111: Since you always use `nr_try`, this could be an iter over `1..=self.tries`

src/scanner/mod.rs
```diff
@@ -100,41 +103,59 @@ impl Scanner {
...
+            debug!("Try number: {}", nr_try + 1);
+
+            ret = match self.connect(socket).await {
```
niklasmohrin, line 114: I think it would be clearer if the `match` did not return anything and you moved this assignment where it actually happens. This way, you could also check whether you really need the assigned value (more in my comment at the end of the function).

src/scanner/mod.rs
```diff
@@ -100,41 +103,59 @@ impl Scanner {
...
+                        }
+                    }
+                    _ => Err(io::Error::new(io::ErrorKind::Other, e.to_string())),
+                },
+            };
 
-                Ok(socket)
+            if ret.is_ok() {
+                debug!("Return ok after {} tries", nr_try + 1);
+                return ret;
```
niklasmohrin, line 152: IMO this code should moved to where `Ok(socket)` is returned

src/scanner/mod.rs
```diff
@@ -100,41 +103,59 @@ impl Scanner {
...
-                        panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
-                    }
-                }
-                _ => Err(io::Error::new(io::ErrorKind::Other, e.to_string())),
-            },
         }
+
+        debug!("Return after {} tries", self.tries);
+
+        return ret;
```
niklasmohrin, line 158: The value of `ret` here will always be the error of the last iteration, because all other values will be overridden again or discarded if a connection succeeds. I think it would be clearer, if you checked whether there will be another iteration in the loop body and if not return the current (and therefore the last) error.

src/scanner/mod.rs
```diff
@@ -100,41 +103,59 @@ impl Scanner {
...
+                            || e.to_string().contains("Network is unreachable")
+                        {
+                            debug!("Socket connect error: {} {}", &e.to_string(), &socket);
+                            Err(io::Error::new(io::ErrorKind::Other, e.to_string()))
+                        } else {
+                            debug!("Socket connect error: {} {}", &e.to_string(), &socket);
+                            panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
+                        }
+                    }
+                    _ => Err(io::Error::new(io::ErrorKind::Other, e.to_string())),
```
niklasmohrin, line 146: The `to_string` method is called many times in this scope. I am not sure if the compiler can optimize this away and only perform one allocation, but I think it would be nicer if you stored the `String` in a local variable and used that.

src/scanner/mod.rs
```diff
@@ -100,41 +103,59 @@ impl Scanner {
...
+                            println!("Open {}", socket.to_string());
+                        } else {
+                            println!("Open {}", socket.to_string().purple());
+                        }
+                    }
+
+                    Ok(socket)
                 }
+                Err(e) => match e.kind() {
+                    ErrorKind::Other => {
```
niklasmohrin, line 135: The return value of this expression would be returned out of the function, if `nr_try` is equal to `self.tries`, eliminating the need of the `ret` variable

niklasmohrin (commented):
Sweet! I made some suggestions that I think would make the code a bit more clear. Especially the semantics around the `ret` variable seem a bit off to me. However, as I haven't contributed much to this repository yet, I do not insist that these changes are implemented and I don't know what is best used here. Anyway, good work and thanks for contributing!

eiffel-fl:
You are welcome and thank you for you review!
About the `ret` variable, I am more used to C so I sometimes use C habit in other languages particularly when I do not know well like Rust.
I will try to remove it though but there are some points that I did not understand in your reviews.

src/scanner/mod.rs
eiffel-fl: I corrected it on my side.

src/scanner/mod.rs
eiffel-fl: If `self.tries` is 0 it means the user gives 0 so I can modify the value when assigning it to `Scanner` and printing an information message.

src/scanner/mod.rs
eiffel-fl: I removed the `if` and put the body in place of where `Ok(socket)` was returned.

src/scanner/mod.rs
eiffel-fl: I did not see this problem when I was coding, I corrected it on my side.

src/scanner/mod.rs
eiffel-fl: If I understand correctly you proposed to write something like:
```Rust
Err(e) => {
  match e.kind() {
    ...
  }
  if nr_try == self.tries {
    return e;
  }
}
```
I do not really stand for this solution because it adds an `if` inside the loop which already contains an `if`.
I rather write something like this:
```Rust
for ... {
  ... {
    Ok(x) => {
    }
    Err(e) => {
      ret = e;
      match e.kind() => {
        ...
      }
    }
  }
}
return ret;
```
Since the code after the loop means the loop is finished an so code tries `self.tries` which is a problem.

src/scanner/mod.rs
eiffel-fl: See my twice above comment :).

src/scanner/mod.rs
niklasmohrin: Even if this method never gets called with `self.tries == 0`, there has to be a return value for when it _would_ be 0. Since the error return type contains a `String` and it is possibe that you never enter the loop,  you will have to have some code that creates a string outside the loop. My point is to have that allocation be after the loop, because it is not needed otherwise.

src/scanner/mod.rs
niklasmohrin: Okay sure, this is probably more about style than anything else. As I see it, there is a local variable that gets reassigned in every iteration of the loop and every value except for the last one is discarded in the next iteration anyway, so there is no reason to keep any of them around. Logically, the function returns either an `Ok` with the socket, or it returns the error message from the last try. I think that the `if` versions better reflects this logic, as the last error directly comes from the last connection attempt and is returned without another indirection. With the `if`, the code that handles connection errors and translates them into the API of the function is in one place and it is more clear what is happening.

Although I think that the extra `if` adds less complexity than the extra variable, I could see that one would be unhappy with the code. If the `ret` variable sticks around, here is what I propose: I think something like `last_error` or so would be a more fitting name, because that's what it is effectively. Furthermore, the type of it should IMO be `Option<io::Error>` (initially `None`), so that you don't have to allocate the default error message if you go into the loop. Lastly, the assignment should move into the `Err` match arm (as you already did in the example above), because currently `ret` only holds `Err` variants anyway.

Sorry for the big comment, but I think the current code around `ret` really is more complicated than it has to be. That said, I would be fine with the second approach if other reviewers give it a pass.

niklasmohrin:
Don't get me wrong, it is a valid way to solve the problem and you did a great job overall, but I personally find this pattern very iritating as it is more complex than what it accomplishes (see that one (somehwat long) comment where I try to explain my point).

eiffel-fl:
There is not problem at all.
We are here to discuss and find the best solution, moreover I do agree that my solution makes the code less readable and I also have to adapt to the language.
If I understand correctly Rust philosophy the less `mut` variable you have the happier you are.

src/scanner/mod.rs
eiffel-fl: `cargo build` throws me an error if I did not set `ret` variable so I corrected this (even though in this case if we arrive after the loop the variable would have been set into the loop).

src/scanner/mod.rs
eiffel-fl: I will go with the `if` inside the loop.
I do not think adding a branch will have a great impact on performance and I agree on the fact that the code is more readable with it.

bergabman:
Awesome PR! We are in the need of this functionality so it's a great PR to have. I think I know what @niklasmohrin tries to explain and I have to agree, so I downloaded this branch to test it. It does what it supposed to do, but that error on the top of the function does look a little weird. I think it also costs some memory allocation and reallocation because of mutability and reassignements. Based on your PR, there were just a few things I had to change to avoid extra allocations, and only allocate that error when and if it's needed. Please check the code I sketched up @eiffel-fl @niklasmohrin and let me know what you think. This would be a bit more rustacian style I think :)

```rust
    async fn scan_socket(&self, socket: SocketAddr) -> io::Result<SocketAddr> {
        for nr_try in 0..self.tries {
            match self.connect(socket).await {
                Ok(x) => {
                    debug!(
                        "Connection was successful, shutting down stream {}",
                        &socket
                    );
                    match x.shutdown(Shutdown::Both) {
                        Err(e) => debug!("Shutdown stream error {}", &e),
                        _ => {}
                    }
                    if !self.greppable {
                        if self.accessible {
                            println!("Open {}", socket.to_string());
                        } else {
                            println!("Open {}", socket.to_string().purple());
                        }
                    }
                    return Ok(socket);
                }
                Err(e) => {
                    if nr_try < self.tries {
                        continue;
                    } else {
                        match e.kind() {
                            ErrorKind::Other => {
                                if e.to_string().contains("No route to host")
                                    || e.to_string().contains("Network is unreachable")
                                {
                                    debug!("Socket connect error: {} {}", &e.to_string(), &socket);
                                    return Err(e);
                                } else {
                                    debug!("Socket connect error: {} {}", &e.to_string(), &socket);
                                    panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
                                }
                            }
                            _ => return Err(e),
                        }
                    }
                }
            }
        }
        Err(io::Error::new(
            io::ErrorKind::Other,
            format!("Did not connect to {}", &socket),
        ))
    }
```

One interesting thing, maybe there is a problem with this logic, because no matter what I tried and made sure to cover all the arms inside the loop to or return or continue, the compiler kept complaining that not all the arms return a `Result`. So I added the last line as `Err` and it gets hit with the default `tries`. So it's something I don't get right now. Someone?

niklasmohrin:
@bergabman You need to add semicolons after your match blocks (both the outer one and the one over `e.kind()`). The logic is not right though, you still want to look at the error in every iteration and panic if needed. The check should only include the return statements. This is roughly how I imagine it:

```rust
async fn scan_socket(&self, socket: SocketAddr) -> io::Result<SocketAddr> {
    for nr_try in 1..=self.tries {
        match self.connect(socket).await {
            Ok(x) => {
                debug!(
                    "Connection was successful, shutting down stream {}",
                    &socket
                );
                match x.shutdown(Shutdown::Both) {
                    Err(e) => debug!("Shutdown stream error {}", &e),
                    _ => {}
                }
                if !self.greppable {
                    if self.accessible {
                        println!("Open {}", socket.to_string());
                    } else {
                        println!("Open {}", socket.to_string().purple());
                    }
                }
                return Ok(socket);
            }
            Err(e) => {
                if e.kind() == ErrorKind::Other {
                    if e.to_string().contains("No route to host")
                            || e.to_string().contains("Network is unreachable") {
                        debug!("Socket connect error: {} {}", &e.to_string(), &socket);
                    } else {
                        debug!("Socket connect error: {} {}", &e.to_string(), &socket);
                        panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
                    }
                }
                
                if nr_try == self.tries {
                    return Err(io::Error::new(io::ErrorKind::Other, e.to_string()));
                }
            }
        };
    }
    
    // Only runs if self.tries is 0
    Err(io::Error::new(
        io::ErrorKind::Other,
        format!("Did not connect to {}", &socket),
    ))
}
```

Note that this does not include the caching of the result of `.to_string()`.

bergabman:
I see. I was not sure in the 0..self.tries part but you are right, as we already start with 1 we can use that and inclusive `..=` range so it's easier to compare. It's better. But where I match `Error::kind`, I think I cover everything, as all the errors have kind and if it's not `Other`, it should hit `_`. But if I always continue, that doesn't really matter :). 
How you wrote it is the way to catch the `too many files open` exception and `panic` on it right away (current logic), instead of looping further. It would only cost 2 more loop iterations and it would probably hit panic, but your way seems the most optimal/efficient. 👍

eiffel-fl:
I updated my code to take into account all reviews and discussions!
So please, let me know if there is a problem.

Moreover, can someone reviews the code about reading into the `config` I am not sure if `tries` is correctly taken into account.

I also add a print with `eprintln` to indicate to user that  he/she set `tries` to 0 so the code set it to 1.
I would have liked to use `warning!` from `tui` but I had troubles with rust import system.

src/scanner/mod.rs
```diff
@@ -87,54 +98,72 @@ impl Scanner {
...
+                        } else {
+                            println!("Open {}", socket.to_string().purple());
+                        }
+                    }
+
+                    debug!("Return Ok after {} tries", nr_try);
+                    return Ok(socket);
                 }
+                Err(e) => {
+                    let error = match e.kind() {
```
niklasmohrin, line 140: This is already way nicer, but I think it can be even more tight. This is what I propose:

```rust
let error_string = e.to_string();
if e.kind() == ErrorKind::Other {
    debug!("Socket connect error: {} {}", &error_string, &socket);
    if !error_string.contains("No route to host")
        && !error_string.contains("Network is unreachable")
    {
        panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
    }
}

if nr_try == self.tries {
    debug!("Return Err after {} tries", self.tries);
    return Err(io::Error::new(io::ErrorKind::Other, error_string));
}
```

This avoids the creation a new `io::Error` if it is not needed

src/scanner/mod.rs
```diff
@@ -87,54 +98,72 @@ impl Scanner {
...
     /// Deals with the <result> type
     /// If it experiences error ErrorKind::Other then too many files are open and it Panics!
-    /// ese any other error, it returns the error in Result as a string
-    /// If no  errors occur, it returns the port number in Result to signify the port is open.
+    /// If any other error, it returns the error in Result as a string
+    /// If no errors occur, it returns the port number in Result to signify the port is open.
     /// This function mainly deals with the logic of Results handling.
     /// # Example
     ///
     ///     self.scan_port(10:u16)
```
niklasmohrin, line 110: This does not look right, what's going on here?

src/scanner/mod.rs
```diff
@@ -36,13 +40,20 @@ impl Scanner {
...
         greppable: bool,
         port_strategy: PortStrategy,
         accessible: bool,
     ) -> Self {
         Self {
             batch_size,
             timeout,
+            tries: if tries == 0 {
+                eprintln!("tries is 0, corrected to 1!");
+                tries + 1
```
niklasmohrin, line 53: Why not just `1`? By the way, if you disallow zero here, I think [`NonZeroU8`](https://doc.rust-lang.org/std/num/struct.NonZeroU8.html) would make a great API and it would remove that last part in the `scan_socket` function. I would still keep the constructor argument as `u8` though and correct in here like you do.

niklasmohrin (commented):
Awesome, we are getting there! Apart from a small suggestion from `cargo clippy`, the `Ok` arm looks perfect to me. I suggested a refactor in the `Err` arm, but it's mostly a syntax change with a slight performance improvement, as the logic was impeccable ✨ .

While reading your code that corrects the value of `tries` to 1 if 0 was given, I remembered that there is [`NonZeroU8`](https://doc.rust-lang.org/std/num/struct.NonZeroU8.html) etc. If you were to use that, you could get rid of the `Err` after the for loop, because then you know that there will be at least one iteration (and that the last iteration will return). Note that the compiler might not figure that out by itself. If you get an error about a missing return after the loop, you just put `unreachable!()` after the loop to tell the compiler just that.

I have not come around to look at the config part yet, because I will have to read some documentation for that too and it is getting late. Maybe I will review tests and config tomorrow.

That said, great work and thank you for being so cooperative! I hope you don't take my large requests amiss and keep on writing great (Rust) code! 🦀

src/scanner/mod.rs
niklasmohrin: This comment is targetted only at the example given here

bergabman:
That panic is a crash, but not as serious as a kernel panic for sure :D. We will handle that logic different in the future.
I also agree with that if statement there, so I think the structure/logic is good like this!

Those macros are exported to the root of the crate, so you can use them with 
```rust
use super::warning;
```
and then 
```rust
warning!("somestring");
```
To get Options and Config merged right, please add the tries to the input.rs merge_required! macro.
`merge_required!(addresses, greppable, accessible, batch_size, tries, timeout, scan_order, command);`

I will ask @bernardoamc to take a look at it as well, but I also think we are getting there and it will be ok for a merge. And we have to change the README and the manual so people will know about the function. :D

src/input.rs
```diff
@@ -95,6 +95,10 @@ pub struct Opts {
     #[structopt(short, long, default_value = "1500")]
     pub timeout: u32,
 
+    /// The number of tries before a port is assumed to be closed.
+    #[structopt(long, default_value = "1")]
+    pub tries: u8,
```
bernardoamc, line 100: We also need add it to the line below so we can infer this information from the config file. :) 

 https://github.com/RustScan/RustScan/blob/5b35a09b0b449d77106668746a1a0b3a8c7d2c5e/src/input.rs#L156

bernardoamc (commented):
First thank you so much for the contribution. <3 

I have a proposal that might change the code a bit.

**Proposal:**

I think we could simplify a lot this logic by just retrying the scanner itself inside `main.rs`. It would go like this:

1. Create a set inside `main`. 
2. Run the scanner
3. Store open ports within set
4. Go to step 1 until retries are exhausted.

**Benefits:**

1. Keep the Scanner with a single responsibility
2. Small change required, so the code becomes more intuitive

**Downsides**

1. We might scan open ports multiple times

I don't mind the downside that much since usually the majority of ports are closed, so we will scanner pretty much the same number of ports during every iteration.

cc/ @bergabman @niklasmohrin @eiffel-fl

src/scanner/mod.rs
eiffel-fl: I corrected it on my side.
About `NonZeroU8` why not, but I prefer correcting it for the user instead of exploding because it is impossible to set a `NonZeroU8` to 0.

src/input.rs
eiffel-fl: @bergabman I add this on my local version.

eiffel-fl:
@bergabman  Thank you for the tip to include `warning!` I try a lot of combination but not this simple one...
I also modify the `README.md` on my local version, but where is the manual you are talking about? So I will update it too.

@bernardoamc If the downside is not so important why not.
I did not go to a solution like this when I implemented the first version because I though `RustScan` goal was to be as performant as possible.
I will let others express their opinion about your suggestion.

src/scanner/mod.rs
eiffel-fl: I did not understand your comment.
But I corrected `ese` to `else` on my side and remove the double space.

niklasmohrin: No no, your diff is fine, I meant the example that is here. It says `self.scan_port(10:u16)`, but that's not the method name and I don't know the `10:u16` syntax.

src/scanner/mod.rs
niklasmohrin: No, I meant if you were to make `self.tries` a `NonZeroU8`, you could have that guarantee in the `scan_socket` method. And of course, you would still correct the 0 here

niklasmohrin:
@bernardoamc Wow, that approach is tempting indeed, this would be a very simple addition :sweat_smile:

However, I would prefer the current approach as it is (probably) more performant and the condition that determines "port open" vs. "port closed" is in one place.

Furthermore, it allows for more specialized additions down the road (for example different number of tries allowed for different IPs or ports). Maybe we should consider refactoring the `scan_socket` method to not take `&self` (so we make it static) and instead have everything needed given to it as an argument (that would be `tries` and `timeout`). This would allow for other methods to use `scan_socket` for other strategies or so.

Yet, I can see how the easier change may be more appropriate here. I am fine with both versions, but I personally prefer the current approach.

bernardoamc:
@niklasmohrin I'm fine with the current approach too, and when the need to further specialize this logic comes we can extract this logic to a self contained struct. Something like a "SocketStrategy" of sorts.

Most of my comments were around style, but @niklasmohrin and @bergabman did a great job reviewing the PR already. As soon as they are adjusted we can ship this great addition 🎉

src/scanner/mod.rs
eiffel-fl: I add this on my local version.

src/scanner/mod.rs
eiffel-fl: I updated the documentation to follow the function name and argument on my local side.

src/scanner/mod.rs
```diff
@@ -87,54 +98,72 @@ impl Scanner {
...
+                    return Ok(socket);
                 }
+                Err(e) => {
+                    let error = match e.kind() {
+                        ErrorKind::Other => {
+                            let error_string = e.to_string();
+                            if error_string.contains("No route to host")
+                                || error_string.contains("Network is unreachable")
+                            {
+                                debug!("Socket connect error: {} {}", &e.to_string(), &socket);
```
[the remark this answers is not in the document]
eiffel-fl: I forgot to update this when I add `error_string` variable.
This is now corrected on my local version.

src/scanner/mod.rs
```diff
@@ -87,54 +98,72 @@ impl Scanner {
...
+                                Err(io::Error::new(io::ErrorKind::Other, e.to_string()))
+                            } else {
+                                debug!("Socket connect error: {} {}", &e.to_string(), &socket);
```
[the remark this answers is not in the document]
eiffel-fl: Code factoring is always cool! This is now the code on my local version.

src/scanner/mod.rs
eiffel-fl: I switch to `NonZeroU8` for `tries` inside `Scanner` on other locations (*e.g.* `input.rs`) `tries` is still an `u8`.

eiffel-fl:
I added all the reviews but stick with the implementation where `tries` takes place inside `scan_socket`.
Let me know if there is a problem or I I can improve something!

src/scanner/mod.rs
```diff
@@ -36,13 +40,20 @@ impl Scanner {
...
     ) -> Self {
         Self {
             batch_size,
             timeout,
+            tries: if tries == 0 {
+                warning!("tries is 0, corrected to 1!");
+                NonZeroU8::new(1).unwrap()
+            } else {
+                NonZeroU8::new(tries).expect("tries cannot be 0!")
+            },
```
bernardoamc, line 56: Here we can do:

```rust
tries:  NonZeroU8::new(tries).unwrap_or(1)
```

I wouldn't bother adding a warn in this case. :)

src/scanner/mod.rs
```diff
@@ -87,54 +98,71 @@ impl Scanner {
...
-                match x.shutdown(Shutdown::Both) {
-                    Err(e) => debug!("Shutdown stream error {}", &e),
-                    _ => {}
-                }
-                if !self.greppable {
-                    if self.accessible {
-                        println!("Open {}", socket.to_string());
-                    } else {
-                        println!("Open {}", socket.to_string().purple());
+        let tries = self.tries.get();
```
bernardoamc, line 114: Following [my last comment ](https://github.com/RustScan/RustScan/pull/266/files#r505692887), now we can use tries normally, so this would become `let tries = self.tries;`

src/scanner/mod.rs
```diff
@@ -11,9 +11,12 @@ use futures::stream::FuturesUnordered;
 use std::{
     io::ErrorKind,
     net::{IpAddr, Shutdown, SocketAddr},
+    num::NonZeroU8,
     time::Duration,
 };
 
+use super::warning;
```
niklasmohrin, line 18: Please merge this with `use super::PortStrategy;`

src/scanner/mod.rs
niklasmohrin: Almost, but the value in `unwrap_or` needs to be `NonZeroU8` too:
```rust
tries: NonZeroU8::new(tries).unwrap_or_else(|| NonZeroU8::new(1).unwrap()),
```

src/scanner/mod.rs
```diff
@@ -87,54 +98,71 @@ impl Scanner {
...
+                        if !error_string.contains("No route to host")
+                            && !error_string.contains("Network is unreachable")
+                        {
+                            panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
+                        }
+                    }
+
+                    if nr_try == tries {
+                        debug!("Return Err after {} tries", tries);
+                        return Err(io::Error::new(io::ErrorKind::Other, e.to_string()));
```
niklasmohrin, line 157: You can use `error_string` here

src/scanner/mod.rs
```diff
@@ -87,54 +98,71 @@ impl Scanner {
...
                 }
-                _ => Err(io::Error::new(io::ErrorKind::Other, e.to_string())),
-            },
+            };
         }
+        return Err(io::Error::new(
+            io::ErrorKind::Other,
+            format!("self.tries is {} so port scanning is not done!", tries),
+        ));
     }
```
niklasmohrin, line 166: Now that `self.tries` is non-zero, this code never runs and can be replaced with just `unreachable!();`.

src/scanner/mod.rs
eiffel-fl: I had trouble with this solution.
Indeed, when I tested it locally I got the following error messages from compiler:
`
tries: NonZeroU8::new(tries).unwrap_or(1),
   |                                                    ^ expected struct `std::num::NonZeroU8`, found integer
...
tries: NonZeroU8::new(tries).unwrap_or(1),
   |                                                    ^ expected struct `std::num::NonZeroU8`, found integer
`

I solved that by writing:
```rust
tries: NonZeroU8::new(tries).unwrap_or(NonZeroU8::new(1).unwrap()),
```
But I find this solution heavy... But maybe I am doing it wrong an there is a lighter solution.

Moreover I think it is important to inform the user that the code modified a parameter he/she gave on the CLI.

niklasmohrin (commented):
Nice! I think it is almost perfect now. Additionally to these suggestions, the suggested refactoring on line 128 is not in yet.

Note: I think you should be able to close older suggestions with a button saying `resolve conversation` to make the page here a bit shorter :smile:

src/scanner/mod.rs
eiffel-fl: I added it locally.

src/scanner/mod.rs
niklasmohrin: It does not really get lighter than that. The shortest I could think of is

```rust
use std::cmp;
// ...
  tries: NonZeroU8::new(cmp::max(tries, 1)).unwrap(),
```

src/scanner/mod.rs
eiffel-fl: For the `unwrap_or`, see my [above comment](https://github.com/RustScan/RustScan/pull/266#discussion_r505733950).

For the `error_string`, next time I use the regex of my editor to replace them all!

src/scanner/mod.rs
eiffel-fl: I had problem if I did not let something after the loop, but replacing it with `unreachable!();` does the job.
`Rust` compiler should be young to not detecting this code is unreachable.

src/scanner/mod.rs
```diff
@@ -87,54 +98,72 @@ impl Scanner {
...
+            match self.connect(socket).await {
+                Ok(x) => {
+                    debug!(
+                        "Connection was successful, shutting down stream {}",
+                        &socket
+                    );
+                    match x.shutdown(Shutdown::Both) {
+                        Err(e) => debug!("Shutdown stream error {}", &e),
+                        _ => {}
                     }
```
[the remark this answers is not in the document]
eiffel-fl: I forgot to take into account this review...
Normally I added it on my local version, but I am not really sure if I wrote it correctly, can you please check it?

eiffel-fl:
Normally I added everything!
The button `resolve conversation` was in front of me and I did not see it :rofl::rofl::rofl:!

src/scanner/mod.rs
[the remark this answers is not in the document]
niklasmohrin: Yes, that's how it's done!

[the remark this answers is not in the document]
eiffel-fl: Perfect!

src/scanner/mod.rs
eiffel-fl: Why not, but I think we need to discuss about informing or not the user?
What do you think about it?

bernardoamc: In my opinion I wouldn't worry about informing the user here, someone has to actively pass `0` to `tries`. It's more of a matter of protecting execution than user experience at this point.

niklasmohrin: I am not really sure what's best here, so I am going to side with @bernardoamc. However, you could mention that `tries = 0` will be automatically changed to 1 in the manual page and maybe even in the help message.

bergabman: I just tested it and it runs fine, from commandline as well as from the config. The number gets changed to 1 if it's 0 and the warning pops up. And this is where I made a mistake by not explaining good enough how to use that macro. :/
We have an accessibility option, where people can chose to have the RustScan output in a screen-reader format, without ASCII art. Could you please change that macro by adding the 2 keyword that's already there in the context, so it gets printed with those options. 
```
warning!("tries is 0, corrected to 1!", greppable, accessible);
```

src/scanner/mod.rs
```diff
@@ -87,54 +96,67 @@ impl Scanner {
...
+
+                        if !error_string.contains("No route to host")
+                            && !error_string.contains("Network is unreachable")
+                        {
+                            panic!("Too many open files. Please reduce batch size. The default is 5000. Try -b 2500.");
+                        }
+                    }
+
+                    if nr_try == tries {
+                        debug!("Return Err after {} tries", tries);
```
bergabman, line 153: Please remove this `debug`, with this the debug output gets flooded with connection errors in debug mode :)

bergabman:
I tested it and made 2 small comments so the output gets improved a bit. Besides this I think we discussed the important parts and I'm happy with the final code. Thanks again for this great PR!

src/scanner/mod.rs
eiffel-fl: I removed it from my local version.

src/scanner/mod.rs
eiffel-fl: I went with an `unwrap/cmp`-based solution and updated `README.md` as well as `--help` message.

eiffel-fl:
@bergabman You are welcome! And thanks also the others for your reviews!
I added them and push the new code.

src/scanner/mod.rs
```diff
@@ -87,54 +91,66 @@ impl Scanner {
...
+                    if let Err(e) = x.shutdown(Shutdown::Both) {
+                        debug!("Shutdown stream error {}", &e);
+                    }
+                    if !self.greppable {
+                        if self.accessible {
+                            println!("Open {}", socket.to_string());
+                        } else {
+                            println!("Open {}", socket.to_string().purple());
+                        }
                     }
```
bee-san, line 129: This should be replaced with the TUI, see documentation here:
https://github.com/RustScan/RustScan/wiki/Contributing-Tips#outputting-information
and code here:
https://github.com/RustScan/RustScan/blob/master/src/tui.rs

bee-san (changes requested):

src/scanner/mod.rs
bee-san: Actually, it shouldn't be replaced -- my bad

bee-san (approved):

niklasmohrin:
Hey, I think I found a neat little adjustment: Since `NonZeroU8` implements `FromStr`, you can actually require the flag to be non-zero all the way from the start. That means the user gets an explicit error instead of a silent one:

 ```console
$ target/debug/rustscan --tries 0 192.168.178.97 --no-nmap
error: Invalid value for '--tries <tries>': number would be zero for non-zero type
```

All I did was change `u8` to `NonZeroU8` in `Config`, `Scanner` etc. This also gets rid of the conversion in `Scanner::new`.

bergabman (approved):
Thanks again for the PR. Great code!

bee-san:
@all-contributors please add @eiffel-fl  for code
@all-contributors please add @niklasmohrin for code

bee-san:
@all-contributors please add @eiffel-fl for code

bee-san:
@all-contributors please add @niklasmohrin for code

niklasmohrin:
Uhhhm, @bee-san I actually only reviewed this PR 😄 I opened #277 yesterday, so I guess I am going to make up for the "code-contributor" title soon 😅

bee-san:
> Uhhhm, @bee-san I actually only reviewed this PR smile I opened #277 yesterday, so I guess I am going to make up for the "code-contributor" title soon sweat_smile

ahhh my bad, I saw your name in the reviews a lot so I thought I'd give it to ya :laughing:

eiffel-fl:
Thank you for the merge!
Good language, good project and good community :smile:!

PR #271, opened 2020-10-12
Make address an optional instead of a positional argument

### LINKED ISSUE #211: Adjust `-p` argument with a single port.
## Bug

Right now the following command will fail: `-b 100 -p 80 127.0.0.1`.

The current alternatives are:

* `-b 100 -p 80, 127.0.0.1`
* `-p 80 -b 100 127.0.0.1`

bernardoamc:
This has the benefit of making the input flexible and fixing an existing bug that happens when a single port is specified, see #211.

Other benefits:
  - We are now parsing only failures as a file, reducing the amount of
    attempts to open a file.
  - We check if a file exists before trying to open it.

⚠️ **This is a breaking change in our API** ⚠️ 

Fixes #211

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,6 +1,7 @@
 use serde_derive::Deserialize;
 use std::collections::HashMap;
 use std::fs;
+
 use structopt::{clap::arg_enum, StructOpt};
 
 const LOWEST_PORT_NUMBER: u16 = 1;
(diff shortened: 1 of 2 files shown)


bee-san:
I'll tag this as 2.0.0 if that's okay with you, as I see this as a major breaking change to anyone's current workflows :)

bernardoamc:
Sounds like a good plan. :)

bee-san (approved):

PR #287, opened 2020-10-17
Add clippy linting and fix existing warnings

### LINKED ISSUE #276: Linting with `cargo clippy`
Hi everyone!

I saw that you guys already have `cargo fmt` set up as a workflow, which is great. I think the addition of linting as a Github workflow would be very nice and help even more to keep code quality up. Currently, there are some warnings when running `cargo clippy` that need to be fixed, but it shouldn't take very long. A workflow could be adapted from [the tealdeer repository](https://github.com/dbrgn/tealdeer/blob/151d014f5d0e48b6988774b9d41d0e8f241c510b/.github/workflows/ci.yml#L41).

If you guys agree, I could open a PR myself and fix the few warnings to get this started.

niklasmohrin:
Closes #276

I split the stuff up into separate commits, but I guess they _could_ also be squashed.

- Fix clippy warnings around match expressions
- Fix clippy warnings about syntax
- Fix clippy warnings around return statements that could just be expressions
- Fix clippy warnings around under-specific method calls
- Add clippy workflow

I am not sure about the workflow as it uses some "secret token", but this is how it is set up over at `tealdeer`. On the project page, it says that it is used to add the annotations to the PR diff. If you don't like that, I could replace it with a workflow that just checks the lints.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -47,6 +47,21 @@ jobs:
           command: fmt
           args: --all -- --check
 
+  clippy:
+    name: Clippy linting
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+            toolchain: stable
+            components: clippy
+            override: true
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-features
+
   links:
         name: markdown-link-check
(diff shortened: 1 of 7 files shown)


niklasmohrin:
You can already see the workflow in action. I think the lints should be required to pass, but I left it optional for now. Let me know what you think!

bernardoamc (approved):
Awesome <3

bee-san:
Amazing!!! Thank you so much!! We've been meaning to do this for a while :P @all-contributors please add @niklasmohrin for code and bugs!

PR #297, opened 2020-10-19
Print seconds elapsed since beginning of scan when an open port is found

### LINKED ISSUE #212: Include execution time in feedback/output
When running a typical scan, the program will provide some stdout as it progresses.  It would be nice to see the total time lapsed since it started, in this output.

eiffel-fl:
Hi!

I hope you are fine and the same for your relatives!

i liked contributing RustScan last time, so I wrote small code to fixes #212.
Please let me know if there any problem with the code or if we can improve it!

Best regards.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -110,6 +110,11 @@ pub struct Opts {
     #[structopt(long, default_value = "1")]
     pub tries: u8,
 
+    /// The period at which rustscan informs about ports during scanning.
+    /// If set to 0, rustscan will not inform about ports during scanning.
+    #[structopt(long, default_value = "60")]
+    pub time_interval: u64,
+
     /// Automatically ups the ULIMIT with the value you provided.
     #[structopt(short, long)]
     pub ulimit: Option<rlimit::RawRlim>,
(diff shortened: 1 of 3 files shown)


bee-san:
I actually have mixed feelings about this, for A11Y purposes and general mess. Thanks so much for contributing this code! I'll think on it, in the meantime I'll grant you the hacktoberfest tag :)

eiffel-fl:
I agree with you on the fact that it adds verbosity to RustScan and maybe it is too much.
Moreover, and as you pointed, we need to take into account accessibility so RustScan can be used by anyone.
Take the time to think about it and if you have an idea of contribution I can try to implement it!

bergabman:
We could keep the accessibility as it is now by removing the `[{}s] ` from [this](https://github.com/RustScan/RustScan/pull/297/files#diff-5d8a67adb5c06f47c290425b99df374d308d6d991de871f68396b5dc4b51aec8R124l) line.

But I have some other problem with this way of runtime measurement. With this change we would get a elapsed time since the beginning of the scan every time if there is an open port found. But what if there is no open port? Or huge gaps between finding an open port? I think the best option would be to have an independent thread(?) or async task(?) that tracks the elapsed time and can inform the user in predefined time intervals, like every 30 sec or 1 minute. That way we would have full control on the measurement, possibly could create an `Opts` for it, and could also avoid the allocation of `Instant::now()` instances when creating a socket. What's your opinion about this?

eiffel-fl:
@bergabman You are right that if there is no open port found the user does not have information about elapsed time.
I can try to implement such a thing, I never wrote thread or async task in rust so I will discover how to do it.
Unfortunately I can not give an ETA for this but I will try to write it :)

22f-cy (approved):

PR #293, opened 2020-10-18
Redesign of README

bee-san:
What do you think? :) 

See the rendered markdown version here:
https://github.com/RustScan/RustScan/blob/1ccd59431e02f19ab2c0212369553307f58c99b3/README.md

ToDo
- [x] Change logo
- [x] Write the documentation on accessibility and RSE (feel free to help with this, any core member can edit docs)
- [x] New terminalizer gifs

PR #298, opened 2020-10-19
Add some instructions to create contributing development environment.

eiffel-fl:
Hi!

When I first contributed to RustScan I used the committed `Dockerfile` to create a "ready to dev" image.
I think this can be useful for anyone so I share it with you!

Best regards.

### DIFF
--- a/contributing.Dockerfile
+++ b/contributing.Dockerfile
@@ -1,0 +1,5 @@
+FROM rust
+# Install nmap first.
+RUN apt-get update -qy && apt-get install -qy nmap
+# Then install rustfmt and clippy for cargo.
+RUN rustup component add rustfmt clippy
--- a/contributing.md
+++ b/contributing.md
@@ -29,3 +29,38 @@ If you have any feature suggestions or bugs, leave a GitHub issue. We welcome an
 
 ## Rewarding you
 I cannot pay you :-( But, I can place your GitHub profile on the README under `#Contributors` as a thank you! :)
+
+## Contributing development environment
+
+To ease contribution to RustScan, you can use the `contributing.Dockerfile` to create a Docker image ready to build and play with RustScan.
+To build it you just need to run:
+
+```bash
+you@home:~/RustScan$ docker build -t rustscan_contributing -f contributing.Dockerfile
+```
+
+Then you need to run the container with a volume so it can access, *with read and write permissions*, to RustScan files:
+
+```bash
+you@home:~/RustScan$ docker run -ti --rm -v "$PWD":/rustscan -w /rustscan rustscan_contributing bash
+```
+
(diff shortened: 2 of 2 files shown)


bee-san:
The README is being completely reworked in #293 , can you copy and paste what you want in a comment, remove it from this commit and I'll add it to the wiki? :D

eiffel-fl:
I saw the README was reworked and the new version seems cool!

The two things that I think can be useful to add to the wiki to use contribution are:
1. This Dockerfile:
```Dockerfile
FROM rust
# Install nmap first.
RUN apt-get update -qy && apt-get install -qy nmap
# Then install rustfmt and clippy for cargo.
RUN rustup component add rustfmt clippy --toolchain 1.46.0-x86_64-unknown-linux-gnu
```
2. And the recipe to use it:
> ## Contributing development environment
> 
> To ease contribution to RustScan, you can use the `contributing.Dockerfile` to create a Docker image ready to build and play with RustScan.
> To build it you just need to run:
> 
> ```bash
> you@home:~/RustScan$ docker build -t rustscan_contributing -f contributing.Dockerfile
> ```
> 
> Then you need to run the container with a volume so it can access, *with read and write permissions*, to RustScan files:
> 
> ```bash
> you@home:~/RustScan$ docker run -ti --rm -v "$PWD":/rustscan -w /rustscan rustscan_contributing bash
> ```
> 
> You can now modify RustScan files with your favorite editor, once you want to compile and test your modifications, type the following in the container prompt:
> 
> ```bash
> root@container:/rustscan# cargo build
> ```
> 
> You are now ready to use RustScan:
> 
> ```bash
> root@container:/rustscan# cargo run -- -b 2000 -t 5000 -a 127.0.0.1
> ```
> 
> You can also format, lint with `clippy` and test the code with the following commands:
> 
> ```bash
> root@container:/rustscan# cargo fmt
> root@container:/rustscan# cargo clippy
> root@container:/rustscan# cargo test
> ```

You can shrink the explanation about using the Dockerfile and running the container/

niklasmohrin:
I think this better fits in the wiki than in the readme and I would rephrase it to suggest Docker as _one possible_ setup. The three `cargo` commands at the end are useful for everyone though and could be mentioned there. Finally, the version number should be replaced with just "stable" in the rustup command (Rust 1.47 is already out!). Other than that, good stuff!

eiffel-fl:
You are right about `docker`, this is one method, not the method.

Can I add this to the wiki? Moreover, I remember that I did not add documentation about `--tries` option in the wiki...

bee-san (approved):

bee-san:
In wiki :) https://github.com/RustScan/RustScan/wiki/Contributing

eiffel-fl:
Thank you for the merge!

Issue #262, opened 2020-10-04
Create a program that when CI runs, runs a binary (HyperFine) and uploads the JSON results to Google Sheets


bee-san:
# Problem

We want to benchmark RustScan with https://github.com/sharkdp/hyperfine  and store the results in Google Sheets.

We want this program (written in whatever language you want, but preferably Python) to run on a new tag release.

You can do this using either GitHub actions or Travis CI.

# What the program should do
1. Download & install HyperFine (this can be done in the CI)
2. Download / Install RustScan (in CI you may want to download Rust, clone the repo, `cargo build --release` to get a binary.
3. Run HyperFine with this binary. Specifically the command is:

```
hyperfine --export-json results.json --warmup 1 '/target/release/rustscan rustscan.cmnatic.co.uk -b 5000 -u 10000 --no-scripts'
```

This runs `rustscan rustscan.cmnatic.co.uk -b 5000 -u 10000` note: you may need to change the batch size if GitHub / Travis freaks out.

It stores it in a JSON file called `results.json` in the current directory.

4. Upload the results to the Google Sheet
Here is how the Google Sheets API works. 
https://developers.google.com/sheets/api/quickstart/python

We **only** want a single column containing the time it took to run.

We suggest you create your own google sheet to test this, before trying to integrate with us.

# Integrating with our google sheets

When you integrate with us, you'll need these secrets:

```
SHEETS_CLIENT_ID
```

```
SHEETS_CLIENT_SECRET
```

![image](https://user-images.githubusercontent.com/10378052/95028022-047b1100-0695-11eb-8166-2162cbe38945.png)

In GitHub CI they are stored as this. **If you used Travis CI, please tell us and we can add them to Travis too**.

We also do not have a specified sheet _yet_. Let us know when you get to that stage and we will create one for you.

Thanks so much! :heart:

CMNatic:
This sounds wicked to me, I'll look into getting google spreadshets as a data source for the grafana whilst anyone/someone works on this (:

bee-san:
> This sounds wicked to me, I'll look into getting google spreadshets as a data source for the grafana whilst anyone/someone works on this (:

Cmnatic is not doing the issue btw, they are doing something else with the google sheet (this issue is still available)

CMNatic:
^ thanks for clarifying 😅

mshaugh:
I've started working on this

PR #312, opened 2020-11-12
Benchmark ci

betta-cyber:
#262 

I create a python program that when CI runs, It can runs HyperFine and uploads the results to Google Sheets

There are some intractable problem when I use google OAuth Credentials. The program need enter some token when i use it on terminal. So I choose Service Account as authorization. You can follow this [doc](https://pygsheets.readthedocs.io/en/stable/authorization.html#service-account) to enable Google Sheets API and generate OAuth Credentials. Download json file.

And then. need add `SHEETS_CLIENT_EMAIL` and `SHEETS_PRIVATE_KEY` in github secrets.

image

You can change the [sheet name](https://github.com/betta-cyber/RustScan/blob/3e76cdb8aca689b7ac26f4b0536b4cee17e89fd9/generate_benchmark.py#L31) in python program. I just upload the time result simply.

The most IMPORTANT things is to add your service_account_email to the share list.

image

I hope that will be helpful.


bee-san:
actually this is quite cool!!

PR #373, opened 2021-03-30
Upgrade Alpine to 3.19.x

### LINKED ISSUE #535: Rustscan cannot saved output anymore with -oN/-oA command
**Describe the bug**
It was working ok till someday it didn't. I didn't know what happened. I was using the same version, the same command, and somehow get this error. Then, I tried to reinstall it again, use different approach such as via docker but still getting the same error.
If I did not use -oN/-oA command, rustscan is working just fine. The problem is that I need the output to be saved.

**To Reproduce**
1. Run the command
```
docker run -it --rm --name rustscan rustscan/rustscan:2.1.1 -a 127.0.0.1 -- -Pn -A -oN test
```
2. See error below
```
$ docker run -it --rm --name rustscan rustscan/rustscan:2.1.1 -a 127.0.0.1 -- -Pn -A -oN test
The Modern Day Port Scanner.
________________________________________
: http://discord.skerritt.blog           :
: https://github.com/RustScan/RustScan :
 --------------------------------------
Real hackers hack time ⌛

[~] The config file is expected to be at "/home/rustscan/.rustscan.toml"
[~] File limit higher than batch size. Can increase speed by increasing batch size '-b 1048476'.
Open 127.0.0.1:1337
[~] Starting Script(s)
[>] Running script "nmap -vvv -p {{port}} {{ip}} -Pn -A -oN test" on ip 127.0.0.1
Depending on the complexity of the script, results may take some time to appear.
Failed to open normal output file test for writing
QUITTING!
[!] Error Exit code = 1
```

**Expected behavior**
I'm expecting to see the normal nmap output without any errors.

gaby:
* Upgrade Alpine from `3.17.x` -> `3.19.x`
* Upgrade nmap to `7.94`
* Add `ca-certificates` and `bind-tools`.
* Add docker and github-actions to dependabot
* Update all the Docker related actions
* Update all the checkout related actions.

Note: Merging this PR will create 8-10 PR's related to Rust outdated actions being used in all the workflows.

Fixes #535


gaby:
Is there anyway to see if there was an improvement in performance by upgrading nmap?

bergabman:
> Is there anyway to see if there was an improvement in performance by upgrading nmap?

You can use `RUST_LOG=info ./rustscan` to see some benchmark data.

bee-san:
@CMNatic thoughts? 🤔

gaby:
@bee-san I will have to update the PR again, it's out of date already 😂

bee-san:
> @bee-san I will have to update the PR again, it's out of date already 😂

hahaha ok <3 :)

gaby:
@CMNatic @bee-san PR has been updated. Also added some other small fixes and fixed the merged conflict.

CMNatic:
Thank you:) reviewing on my end 🎉

CMNatic:
I've tested it manually, and all is good on my end @bee-san. It does not solve #535, but that is because the Dockerfile user (rustscan) does not have permission to write to `/` (the default context), so we will need to modify the Dockerfile to solve that.

gaby:
@CMNatic I found it odd that a non-root user was created for scanning ports 😂

/ ownership could be change, but that could break other things

gaby:
I can remove the rustscan user, mounted directories won't work either unless they are chown/chmod before hand

CMNatic:
> I found it odd that a non-root user was created for scanning ports 😂
Yes, me too 😅 we could do that. No reason perse other than it not being best security practice having the tooling run as root. It would mean that files can be saved anywhere i.e. `docker run -it --rm --name rustscan rustscan -a 127.0.0..1 -- -Pn -A -oN test.txt`, but as you said, mounted directories would need to be chowned and set up.

I think that begins to creep out of the scope for this though / an "advanced" use, similar to bringing your own nmap scripts to run

I'm happy to close and merge with the above :) thanks for the contribution (and the skipping docker image CI check)!

PR #394, opened 2021-06-15
README.md: Do not display <p> on github.io

DiLeGrosPanda:
The README rendered the <p> tags as text without interpreting them as HTML, I removed them and used markdown syntax to center the text.
Added a closing tag and fixed a simple typo as well.
Current problem can be seen there: https://rustscan.github.io/RustScan/

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,10 @@
 </p>
 <hr>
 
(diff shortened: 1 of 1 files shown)


wa7sa34cx:
Nice work! I wanted to do same... Awaiting for merging...

DiLeGrosPanda:
I'm unsure if it was really necessary, the github.io frontpage is different than the github's one so I'm guessing it 'll have to be updated anyway, I don't really know how that work though

PR #403, opened 2021-08-07
Allow the usage of range addresses

patrickelectric:
Like nmap, where I can use it with `192.168.1.2-255 -p 22`


bee-san:
Thanks for this! RustScan already has this feature :) https://github.com/RustScan/RustScan/wiki/Things-you-may-want-to-do-with-RustScan-but-don't-understand-how#ranges-of-ports

Issue #419, opened 2021-12-23
Update readme - Link to docker installation docs doesn't work.


spothound:
**Describe the bug**
![image](https://user-images.githubusercontent.com/15269938/147255136-494ea9f5-1432-4364-b181-8e3f5a9ff6c0.png)
This link doesn't work. Probably the correct one is: https://github.com/RustScan/RustScan/wiki/Installation-Guide#docker-

**To Reproduce**
1. Click on the link shown in the image

**Expected behaviour**
The link should redirect to instructions on how to use the docker container as an executable, probably https://github.com/RustScan/RustScan/wiki/Installation-Guide#docker-

I will open a PR to fix this.

PR #420, opened 2021-12-23
Fix broken link in readme

spothound:
Closes https://github.com/RustScan/RustScan/issues/419


Ccode-lang:
@Thijs0x57 it does not seem to change that link.

PR #455, opened 2022-10-03
fix: crucial easiest way to install rustscan, on Readme add crates

miguelgargallo:
At university some rust fans and I (colleagues) tried to install, but without success, cargo install rustscan

At university some rust fans and I (colleagues) tried to install, but without success, did not work properly with docker, so with this PR we want to encourage new devs to implement with one click install "cargo install rustscan" command, and allow other devs to implement as fast as they can the tool, We think this is an amazing and beautiful tool, and if a team of 6 devs are frustrated to did accomplish using it! We want to add this way, to easy, fast and secure implement 1 command install on machines! Thanks to all team to make projects like this!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,21 +3,6 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.17.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
-dependencies = [
- "gimli",
-]
-
-[[package]]
-name = "adler"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


SPIRY-RO (approved):

README.md
```diff
@@ -189,3 +190,5 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
 <!-- ALL-CONTRIBUTORS-LIST:END -->
 
 This project follows the [all-contributors](https://github.com/all-contributors/all-contributors) specification. Contributions of any kind welcome!
+
+,3
```
bee-san, line 194: This is a lovely PR!

```suggestion
```

Can I suggest removing these two lines? :) <3

bee-san (changes requested):

README.md
miguelgargallo: Yeah let me take a look tomo

miguelgargallo:
here https://github.com/RustScan/RustScan/pull/469

PR #491, opened 2023-01-08
Minor documentation table fix for github.io

lerignoux:
Hello

I noticed the [github io documentation page](https://rustscan.github.io/RustScan/) display some html content. I guess the table embedded html is not supported.

## Changes:
* I added a missing tag.
* I aslo was suggesting to use markdown table columns center features to clean it up.
It should display correctly in the github pages I think.

The only issue is that the last row is centered too unlike in the current readme.

Do you think it is ok ?
If not I can drop the second commit and only fix the missing </p> leaving the github pages soso :P


bee-san:
Looks good to me! Thanks! <3

bee-san (approved):

PR #531, opened 2023-12-25
feat(cli): add custom dns resolver option

khanhnt2:
## What's changed:

1. `trust-dns-resolver` is rebranded to `hickory-resolver` so I updated the dns resolver library to `hickory-resolver`, https://github.com/hickory-dns/hickory-dns/releases/tag/v0.24.0

2. Add `--resolver` option to CLI to use custom dns resolver, you can pass the file path or string comma-delimited list. For example:
```sh
# Download resolvers list to a file
$ curl 'https://raw.githubusercontent.com/trickest/resolvers/main/resolvers-trusted.txt' -o 'resolvers-trusted.txt'
$ rustscan -a hackerone.com --resolver resolvers-trusted.txt
```
Or
```sh
$ rustscan -a hackerone.com --resolver 1.1.1.1,8.8.8.8
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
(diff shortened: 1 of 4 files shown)


bee-san:
Hi! Mind pushing an empty commit? CI has failed to pick this up and I think it'll only pick it up on a new commit? 🤔

khanhnt2:
Hi @bee-san  , I just pushed an empty commit, can you check again?

bee-san:
Thank you @khanhnt2 !

Just some points :)
- [ ] The linting is failing :(

I think this logic might work better:

1. Read from default resolver file. In Linux this is [this file](https://www.baeldung.com/linux/etc-resolv-conf-file). Maybe Hickory has this built in already somewhere or an arg we can use?
^^ This should happen without an argument I think
2. We have a [Rustscan config file ](https://github.com/RustScan/RustScan/wiki/Config-File), I think you can use that here to define resolvers too as it'll be useful to not have to redefine them all the time, and you can have system-wide resolvers vs just Rustscan ones

If user enters the resolver argument:
3. Read from file
4. Split string and use those as resolvers

What do you think? Let me know. Thank you for your contribution! <3 :)

bee-san:
Maybe a bit much for 1 PR.

What about:
1. Add some tests
2. Check if Hickory uses the DNS default file?

And I'll make a ticket to address the others :)

PR #536, opened 2024-02-06
Fixed #527

### LINKED ISSUE #527: RustScan scripts parser error
**Describe the bug**

scripts\mod.rs  ,  line 236
        let arguments = shell_words::split(&to_run).expect("Failed to parse script arguments");

my` to_run` variable like:
python3 C:\\Users\\qwrdxer\\.rustscan_scripts\\test_script.py 192.22.105.149 22,111,25,110,3344,8080,8888

after call the func  ` shell_words::split(&to_run)`
image
the  '\\' was delete !  I don't know more about this lib , here  is my environment:

os: windows10 
ide: Clion

The temporary solution is
change 
`let arguments = shell_words::split(&to_run).expect("Failed to parse script arguments");`
to
`let arguments:Vec<String> =to_run.split_whitespace().map(|s| s.to_string()).collect();`

**screen shot**

image
image

gufengwyx8:
I just removed shell-words but use `Exec::shell` to execute script, then windows script will run successfully.


gufengwyx8 (commented):
fixed

PR #427, opened 2022-01-24
feat(build): add GitHub Action to make release

PsypherPunk:
Add config. to, when a new tag is pushed:

- build a `.deb` file using [`cargo-deb`](https://github.com/kornelski/cargo-deb);
- include the resulting `.deb` file in the subsequent release.

An example of one such release can be seen [here](https://github.com/PsypherPunk/RustScan/releases/tag/v6.6.6).

Relates #83

PR #559, opened 2024-04-09
Update release workflow

PsypherPunk:
Apologies if this is excessive but following #427, there are a few things I noticed:

1. `Cargo.toml` now supports [`strip`](https://doc.rust-lang.org/cargo/reference/profiles.html#strip) which I _think_ should replace the explicit `strip` steps in the release workflow, unless there's any other reason they're not quite identical…?
2. the trigger for the release workflow includes a tag with a `v` prefix. However, no existing tags/releases have that prefix; I've removed it and updated the workflow accordingly.
    - the `docker.yml` has the same trigger—if the above holds true, should that be updated to match?
4. as kinda touched upon in #427, I've made the OS/build checks in each step more generic (i.e. matching `linux`, not `x86_64-linux`) so if future architectures are added, the workflow should accommodate them.

Let me know if I've misread anything or any of the above isn't required/desired.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,7 +60,7 @@ jobs:
 
     steps:
       - name: Set Git config. (windows only)
-        if: matrix.build == 'x86_64-windows' # Windows is missing some dependencies
+        if: contains(matrix.build, 'windows') # Windows is missing some dependencies
         run: |
           git config --global pack.windowMemory "100m"
           git config --global pack.packSizeLimit "100m"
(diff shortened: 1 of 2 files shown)


bee-san:
@PsypherPunk  waiting on @CMNatic  to have a lil look

then i'll merge this and try a release 🥳

PsypherPunk:
@bee-san, come to think of it, I don't think builds triggered by pull-requests get access to secrets (to avoid external parties raising PRs to steal them, presumably): https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

CMNatic:
Hey 👋🏻 

> the docker.yml has the same trigger—if the above holds true, should that be updated to match?

Ahh good catch @PsypherPunk. Yes, I think this should also be updated to match :)

> I don't think builds triggered by pull-requests get access to secrets

This is my understanding as well. I've double-checked the repo secrets for the dockerhub and they are correct. I think the PR will need to be merged for the runner to run.

If it doesn't, I have a PR ready that uses a newer version the Docker login runner, which uses a access token instead that we can test @bee-san

bee-san:
> the docker.yml has the same trigger—if the above holds true, should that be updated to match?

Mind doing this in another PR? :)

PR #560, opened 2024-04-10
Update Docker Runner

CMNatic:
This PR upgrades the DockerHub login job from `v2` to `v3`, as per the workflow's [repository](https://github.com/docker/login-action?tab=readme-ov-file#docker-hub), it recommends using a personal access token and not account password. This should hopefully fix the check failing on PRs (i.e. #559).

I've updated the repository secrets to reference `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` and the `docker.yml` appropriately.

The modification to `main.rs` is just resolving a typo to force the action to be ran.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -25,10 +25,10 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Log in to Docker Hub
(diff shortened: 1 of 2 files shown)


bee-san (approved):
lovely!! <3

PR #561, opened 2024-04-10
build: update Docker workflow trigger

PsypherPunk:
as requested in #559, remove the `v` prefix from the triggering tag for the Docker workflow, to match the standard tagging pattern.


CMNatic (commented):
LGTM

Issue #497, opened 2023-01-24
Release RustScan as a library as well


joojscript:
**Is your feature request related to a problem? Please describe.**
I would love to integrate RustScan inside a Rust application, but currently, the only way to do it is actually installing it on the OS with `cargo install`, and then, calling it by the shell of the system, which is not optimal.

**Describe the solution you'd like**
It would be really great to just add a `lib` entry point in `Cargo.toml` (at least), and release it, so rust developers could embed it into open-source solutions.

**Describe alternatives you've considered**
- Add `lib` entry to `Cargo.toml`;
- Modularize such that RustScan can be used as a library as well;

**Additional context**
N/a

StefanoSetti:
Are you willing to implement it?

bee-san:
I will take any PR that implements RustScan as a library :)

PR #562, opened 2024-04-11
refactor: make RustScan additionally a library

PsypherPunk:
- add `lib.rs` (permitted as per [_Packages and Crates_](https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html)).
- refactor `main.rs` to reference the library.
- update module references as required.

relates #497

By way of example, configuring a new binary project which references this crate:

```toml
[dependencies]
async-std = "^1.7"
rustscan = { path = "/home/rust/git/RustScan" }
```

…allows `rustscan` to be referenced as a library:

```rust
use async_std::task::block_on;
use std::{net::IpAddr, time::Duration};

use rustscan::input::{PortRange, ScanOrder};
use rustscan::port_strategy::PortStrategy;
use rustscan::scanner::Scanner;

fn main() {
    let addrs = vec!["127.0.0.1".parse::<IpAddr>().unwrap()];
    let range = PortRange {
        start: 1,
        end: 1_000,
    };
    let strategy = PortStrategy::pick(&Some(range), None, ScanOrder::Random);
    let scanner = Scanner::new(
        &addrs,
        10,
        Duration::from_millis(100),
        1,
        true,
        strategy,
        true,
        vec![9000],
    );

    let scan_result = block_on(scanner.run());

    println!("{:?}", scan_result);
}
```

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,3 +1,4 @@
+//! Provides a means to read, parse and hold configuration options for scans.
 use serde_derive::Deserialize;
 use std::collections::HashMap;
 use std::fs;
--- a/src/input.rs
+++ b/src/input.rs
@@ -211,6 +211,29 @@ impl Opts {
     }
 }
 
+impl Default for Opts {
+    fn default() -> Self {
+        Self {
+            addresses: vec![],
+            ports: None,
+            range: None,
+            greppable: true,
+            batch_size: 0,
+            timeout: 0,
+            tries: 0,
+            ulimit: None,
+            command: vec![],
+            accessible: false,
+            scan_order: ScanOrder::Serial,
+            no_config: true,
+            top: false,
+            scripts: ScriptsRequired::Default,
+            config_path: None,
+            exclude_ports: None,
+        }
+    }
+}
+
 /// Struct used to deserialize the options specified within our config file.
 /// These will be further merged with our command line arguments in order to
 /// generate the final Opts struct.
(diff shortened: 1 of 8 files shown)


src/input.rs
```diff
@@ -161,6 +161,27 @@ impl Opts {
         opts
     }
 
+    pub fn default() -> Self {
```
PsypherPunk, line 164: This is the only major change (as other tests reference this, they no longer work).

PsypherPunk: I could move this to a `impl Default for Opts` which resolves the `clippy` warning, unless there's a better way…?

PsypherPunk:
Also, tests pass _except_ that as this is now a library, doc-tests are running and failing.

bee-san:
Wow this is cool!!!! A long awaited feature. I think some docs on how to use the library would be great too 🥳

src/input.rs
PsypherPunk: Added `impl Default for Opts`.

PsypherPunk:
@bee-san, I've fixed the failing parts:

- fixed the doc-tests in `src/benchmark/mod.rs`.
- changed the indentation in `src/scripts/mod.rs` as this was being interpreted as code—this might need a review as I'm not sure if it looks as intended in the resulting doc. output.
- I've marked a couple in `src/scanner/mod.rs` as `compile_fail`; I don't think there's a particularly easy way to instantiate a `Scanner` for the examples—I suppose things like this are going to be raised if this is to be used as a library; should there be a sensible `Default` for `Scanner`?

Re. the "docs on how to use the library", I'll see what I can do…

bee-san (approved):

bee-san:
Wow! Thank you so much!!

Issue #533, opened 2024-01-16
Add stable tag to docker


Darks1de42:
**Is your feature request related to a problem? Please describe.**
In the manual stands there is the Tag `alpine`, but this does not exist.

**Describe the solution you'd like**
The use of the stable Tag for the latest stable version would be nice. Then you can use an alias which uses always the latest version.

CMNatic:
Ahh that's interesting. I'll publish the `alpine` tag. The repo has the `latest` tag, which should be considered the stable release. However, there's a problem with the CI/CD pipeline in pushing the actual latest stable release to the repo. This is also on my todo.

Thanks :)

bee-san:
Hi there 👋🏻 

As of November 2024 the RustScan team has decided to remove all support for officially maintaining packages other than `Cargo`.

**Why?**

We are an incredibly small team and we found we were working more on fixing packages / the CI for packages rather than making RustScan a great tool.

**How will I use RustScan?**

RustScan is in many, many package managers already:
https://repology.org/project/rustscan/versions 

As well as on others like Snap https://snapcraft.io/rustscan or Docker https://hub.docker.com/search?q=rustscan and we may even be built into some pentesting distros by default.

If your favourite repo does not have a RustScan package, you are free to either make one or request one from them. RustScan is quite popular with thousands of GitHub stars and downloads, and it is incredibly easy to build (git clone && cargo build). We are sure that if we are not already distributed some way, it should not be hard to make a package.

**What about Cargo?**

Cargo will be our only officially supported package, as it's built into Rust by default and it's what our developers use.

If you want to use it, please install Rust and then run `cargo install rustscan`.

PR #570, opened 2024-04-14
Docker Image Checks Rework

CMNatic:
Hello!

I have modified the docker.yml GitHub action that is used to build and publish the Dockerfile to DockerHub. For quite a while, the check would fail on PRs and releases. This is likely due to the action changing/DockerHub API changing since we implemented the workflow on our end.

I have now used a different action and configured the docker.yml workflow to:

1) Run when a new release is made; this will tag it as `rustscan/rustscan:1.11.11` on DockerHub, for example
2) It now successfully runs when a PR is made. However, the Dockerfile is only built, and isn't actually published. Otherwise, this means our DockerHub will be full of tags such as `rustscan/rustscan:pr-400`, but building it is important for CI/CD purposes, and the check will fail if it fails to compile RustScan.

3) It will now run when a merge or commit is made to the `master` branch, this will result in a `latest` tag being published to DockerHub, so people can try out new features/updates without having to wait for an official release.

4) Use a DockerHub access token (repo secret) to login to DockerHub 

image

image

I've tested it on my end with a demo repository and new DockerHub repo: `cmnatic/githubtest`, but we will need to see how it works out for rustscan/rustscan :)

TODO: figure out how we can determine "what is stable" to create a "stable" tag for #533. Maybe this is just a re-tag of the latest release.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,14 +3,14 @@ name: Publish Docker image
 on:
   push:
     branches:
-      - 'master'
+      - 'master' # push on merge/commit to master to create a latest tag
     tags:
       - '[0-9]+.[0-9]+.[0-9]+'
   release:
     types: [published]
   pull_request:
     branches:
-      - master
+      - master # run part as PR checks, but the image will not be published on DockerHub (as this will tag it as say rustscan/rustscan:pr-400)
     paths:
       - 'Dockerfile'
       - 'Cargo.toml'
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -37,7 +37,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
         with:
-          images: cmnatic/githubtest
+          images: rustscan/rustscan
           flavor: latest=true
 
       - name: Build and push Docker image
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,6 +2,8 @@ name: Publish Docker image
 
 on:
   push:
+    branches:
+      - 'master'
     tags:
       - '[0-9]+.[0-9]+.[0-9]+'
   release:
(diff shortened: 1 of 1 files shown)


CMNatic:
Ahhh so it is still having the same error because it's from a fork. https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow.

@bee-san I don't think we'll be able to fix this. We could just remove the Dockerfile check/build on PRs (remove): 
```  pull_request:
    branches:
      - master
```
So that the check never happens here. I mean, we already have other checks that check if RustScan can be compiled so.... in hindsight, it seems a bit redundant that we're  re-checking if it can compile via Dockerfile, we will just need to review any PRs that change the Dockerfile specifically. If RustScan can't compile via the Dockerfile, other checks such as Build/Test Suite will pick this up.

bee-san:
> Ahhh so it is still having the same error because it's from a fork. https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow.
> 
> @bee-san I don't think we'll be able to fix this. We could just remove the Dockerfile check/build on PRs (remove):
> 
> ```
>     branches:
>       - master
> ```
> 
> So that the check never happens here. I mean, we already have other checks that check if RustScan can be compiled so.... in hindsight, it seems a bit redundant that we're re-checking if it can compile via Dockerfile, we will just need to review any PRs that change the Dockerfile specifically. If RustScan can't compile via the Dockerfile, other checks such as Build/Test Suite will pick this up.

Let's remove the check on PRs :)

gaby:
@CMNatic The only way to test PR changes is to have a script or something that you run locally to pull the PR branch and build the image.

While you can publish images from a PR, it really only works for maintainers/contributors of a project, any fork opens a can of worms in terms of permissions/tokens.

In the meantime you could add support for posting the image to ghcr.io (GitHub Container Registry). Feel free to copy my CI from here: https://github.com/serge-chat/serge/blob/main/.github/workflows/docker.yml

Some notes:
- `latest` tag only gets updated if a new Tag/Release is created.
- `main` tag gets updated  everytime something is merged into `main`
- each release triggers semver tags, ex you tagged v1.0.0 this will create:
  - v1.0.0
  - v1.0
  - v1
- By default tags dont have the prefix `v` you will need to add it in the `patterns` section of the yaml.
- this will build the image on each PR, but not publish it.

PsypherPunk:
@bee-san, in light of the recent packaging decisions, is this effectively defunct?

PR #587, opened 2024-05-10
 feat(cli): add custom dns resolver option

PsypherPunk:
Note: this extends #531 (to which I don't have write-access) but retains all the commits therein (all credit to @khanhnt2).

relates #531

As discussed on that PR:

- fixed the linting issues;
- added some tests (not easy as `Resolver` hides a lot of its internal fields so validating the exact type built doesn't appear to be straightforward).
- updated the derivation to be:
    - if `--resolver` is passed or `resolver` is set in config:
        - assume the value is a file and read that;
        - if not, assume it's comma-separated IP addresses;
    - if no argument is passed:
        - [derive from the system config.](https://github.com/hickory-dns/hickory-dns/blob/main/crates/resolver/src/resolver.rs#L113);
        - if that fails, fall back to the previous CloudFlare behvaiour.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -35,20 +35,18 @@ pub fn parse_addresses(input: &Opts) -> Vec<IpAddr> {
         Resolver::new(ResolverConfig::cloudflare_tls(), ResolverOpts::default()).unwrap();
 
     if input.resolver.ne("") {
-        let resolver_ips =
-            if let Ok(r) = read_resolver_from_file(input.resolver.as_str()) {
-                r
-            } else {
-                // Get resolvers from the comma-delimited list string
-                let mut r = Vec::new();
-                input.resolver.split(',').into_iter().for_each(|item| {
-                    match IpAddr::from_str(item) {
-                        Ok(ip) => r.push(ip),
-                        Err(_) => (),
-                    }
-                });
-                r
-            };
+        let resolver_ips = if let Ok(r) = read_resolver_from_file(input.resolver.as_str()) {
+            r
+        } else {
+            // Get resolvers from the comma-delimited list string
+            let mut r = Vec::new();
+            input.resolver.split(',').for_each(|item| {
+                if let Ok(ip) = IpAddr::from_str(item) {
+                    r.push(ip)
+                }
+            });
+            r
+        };
         let mut rc = ResolverConfig::new();
         for ip in resolver_ips {
             rc.add_name_server(NameServerConfig::new(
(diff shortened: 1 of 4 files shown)


PsypherPunk:
Converting to draft while I fix the conflicts…

PsypherPunk:
Resolved the conflicts but seeing a weird build error:

```sh
error: could not copy file from 'C:\Users\runneradmin\.cargo\bin\rustup-init.exe' to 'C:\Users\runneradmin\.cargo\bin\rustup.exe': Access is denied. (os error 5)
Unexpected error attempting to determine if executable file exists 'C:\Users\runneradmin\.cargo\bin\rustup.exe': Error: EPERM: operation not permitted, stat 'C:\Users\runneradmin\.cargo\bin\rustup.exe'
```

src/address.rs
```diff
@@ -204,4 +255,27 @@ mod tests {
...
+    fn resolver_args_google_dns() {
+        let mut opts = Opts::default();
+        // https://developers.google.com/speed/public-dns
+        opts.resolver = Some("8.8.8.8,8.8.4.4".to_owned());
+
+        let resolver = get_resolver(&opts.resolver);
+        let lookup = resolver.lookup_ip("www.example.com.").unwrap();
+
+        assert!(lookup.iter().next().is_some());
+    }
```
bee-san, line 280: I think we try to avoid network tests so you can test on machines without the internet, in this case mocking the test output 🤔 

But I think for the things which require this (Homebrew requires no networked tests) we have our own different tests specific to it

Although generally I regualrly make the argument that you cannot test a network CLI tool without some kind of network 😂  so approved!

bee-san (approved):
Approved!! Thank you so much and thanks @khanhnt2

src/address.rs
PsypherPunk: Yep, I mulled this one over for a while: the only thing I could think would be to wrap the `Resolver` in a "New Type" which could capture the config. sent to it an expose that. That way, we could:

a. better test that each branch of `get_resolver` results in the right DNS resolvers being referenced.
b. potentially derive a `StructOpt` `from_os_str` for it and create it upstream in `Opts`, isolating a bit more from the parsing code.

…but that seemed a bit much after a hefy `rebase` :grin:

bee-san: I'm aware in PyTest that you can add "tags" to tests, so you could create  tests like "pytest --tag online" to test the online features

that way we can keep it to testing offline for locally and online for CI etc....

Not sure if cargo supports that, and again it isn't really a good solution :( 

Jeeze, testing networking tools is hard!

bee-san:
Hey @PsypherPunk  I added you as a maintainer of RustScan! I think you should be able to push to master and run your own CI now? 🤔

PsypherPunk:
Thanks, @bee-san!

shock

Issue #588, opened 2024-05-14
making the switch from using async_std to tokio


vrtgs:
**Is your feature request related to a problem? Please describe.**
Yes, it is related to a performance and ergonomics issue. Currently, RustScan uses async_std for asynchronous programming. While async_std is a great library, it has been observed that Tokio provides better performance and more ergonomic APIs in some cases.

**Describe the solution you'd like**
I would like RustScan to switch from async_std to Tokio for handling asynchronous tasks. This would involve refactoring the existing code that uses async_std to use Tokio instead. The goal is to improve the performance of RustScan and make the internal code more ergonomic (highly opinionated).

**Describe alternatives you've considered**
An alternative would be to continue using async_std and try to optimize the current implementation. However, given the reported performance and ergonomic advantages of Tokio, switching to Tokio seems pretty promising.

**Additional context**
For users who use this crate as a library, users of async std would not be able to use it, although tokio users would benefit greatly as there is no longer a need to startup an invisible background async_std runtime, and given that tokio is much more popular I belive this is worth it

Vrtgs:
Any updates on this?

PR #589, opened 2024-05-14
Switch to tokio, and update dependencies

Vrtgs:
implement the #588 suggestion

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,9 +356,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.8"
+version = "4.5.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "84b3edb18336f4df585bc9aa31dd99c036dfa5dc5e9a2939a722a188f3a8970d"
+checksum = "64acc1846d54c1fe936a78dc189c34e28d3f5afc348403f28ecf53660b9b8462"
 dependencies = [
  "clap_builder",
  "clap_derive",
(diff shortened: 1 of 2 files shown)


bee-san:
Thank you so much for this! We take performance seriously here. I'll have to read up more on Tokio vs STD:ASYNC but I think you're right, Tokio in my opinion would be faster now-a-days :)

Vrtgs:
opps forgot to make a new branch, sorry for the extra pushes

bergabman:
I'd like to add some things regarding the arguments in issue #588. It's the PR related to it, and I figured we could keep the discussion in one place.
First some context; Back in the day of creating rustscan the situation was slightly different. Tokio had a different api, structure, that was inconsistent with the rust ecosystem, and async-std promised a way more consistent, rust std feel for their async implementation. However this has all changed when async await got stabilized and tokio got completely rewritten. Looking at how async-std got no update in the past 2 years, it makes a whole lot more sense now to switch to tokio, so I agree on that point. 

Considering performance, the scan engine was built on top of `FuturesUnordered`, which is a more generic futures implementation and not part of async-std, it's part of the futures crate. The reason for this decision was performance, and to not rely on async-std in performance critical parts of RustScan. As async await was just stabilized back then, we knew things would change, supporting the decision to stick with FuturesUnordered. 
As @bee-san stated, performance is critical. I noticed that you replaced `FuturesUnordered` with `JoinSet` from tokio in the scan engine. [L81](https://github.com/RustScan/RustScan/pull/589/files#diff-5d8a67adb5c06f47c290425b99df374d308d6d991de871f68396b5dc4b51aec8L81) Based on some discussions in the tokio repo https://github.com/tokio-rs/tokio/issues/5564, this decision would set us back with 5-18% in performance in general, because of a design decision in the implementation of `JoinSet`. 
I haven't done a performance comparison myself, but it would be benficial to see if there is actually a performance difference between the 2 implementations in our case. Based on the benchmarks, and the discussion in that issue, I would prefer to stay with `FuturesUnordered`, but if the benchmark shows different, we can go with `JoinSet`, completely remove the futures dependency and use tokio everywhere.

PR #590, opened 2024-05-14
chore: upgrade dependencies

PsypherPunk:
A few direct dependencies have been updated; for the sake of avoiding a larger, future update, move to the latest versions of direct dependencies:

  - update `rand` usage to use `Range` types.
  - update `cidr_utils` to reflect underlying `cidr` crate.


PsypherPunk:
I was looking at this as a precursor for fixing the `atty` dependency which Dependabot has flagged. However, I've seen that #589 has been raised which covers this and more; happy to close this, depending on the outcome of that one.

Vrtgs:
slight comment on this pull  request, if mine doesn't go through make sure to change the gen_range to an exclusive range, because as pointed to by the 0.7.3 docs, [gen_range](https://docs.rs/rand/0.7.3/rand/trait.Rng.html#method.gen_range) used to use an exclusive range, and so so should we in the newer version

bee-san (approved):

bee-san:
thank you!

Issue #22, opened 2020-07-23
UDP Scan


DanNegrea:
More of a feature request: UDP port scanning is taking exponentially longer than TCP port scanning. 
Would your approach help reduce the scan time?

brandonskerritt:
Hey @DanNegrea , it would reduce the time for sure but I don't currently support UDP (I think, anyway, unless the sockets library does and I don't know about it). But it's on the TO-DO list! :D

cocoonkid:
UDP would be a great addition. I root for it here :flags:

cocoonkid:
I am still rooting   flags :}

fropert:
+1

bee-san:
👋🏻 

The sockets library does support UDP:
https://doc.rust-lang.org/std/net/struct.UdpSocket.html

I think the PR would be a case of just doing an `if` statement and either doing TCP or UDP.

I'll timebox it for this weekend, not sure how to test but I'll figure it out :D

BrendanGlancy:
@bee-san I've been working on this, did you ever get anywhere with it?

bee-san:
> @bee-san I've been working on this, did you ever get anywhere with it?

Hey there!

I did not get anywhere with this, feel free to contribute! 🥳

BrendanGlancy:
@bee-san I don't know if we should open a discussion or something, but what I am finding is that UDP scanning is a little bit more complicated than TCP scanning. I can submit a pull request with what I have now that works for like 60-70% of cases and is much faster than nmap -sU.

The part that is difficult on this is that, essentially, even if we did the 20 most common UDP ports and didn't mess with the wait time for the response, which seems to actually matter in a lot of the testing I did, we'd have to scan each port 20 times. Multiple this but 3 or 4 different timeout lengths and ...

Anyway, like I said I'm happy to submit a pull request, as I'd love some criticism on my shitty junior dev Rust code. But the goal of this would be more or less getting some feedback from people more knowledgable than me on the subject on how to improve this.

bee-san:
> @bee-san I don't know if we should open a discussion or something, but what I am finding is that UDP scanning is a little bit more complicated than TCP scanning. I can submit a pull request with what I have now that works for like 60-70% of cases and is much faster than nmap -sU.
> 
[...]

PR #596, opened 2024-05-21
Udp Scanning 

BrendanGlancy:
So I am new to Rust programming and a junior-level developer, at best, in any language.

#22 This is an attempt to add this feature to Rust Scan, but there are definitely some problems. 

From what I have gathered, nmap does a couple of things for a UDP scan; they try different packet types as well as different wait times in waiting for a response from a UDP port. This makes things very slow, as I am sure you have noticed -sU scans take forever

What I am looking for in submitting this pull request is not for it to be merged, but for suggestions from people much smarter than me so I can improve the functionality of this. 

Also for the automatic nmap scan that runs after every rustscan, I imagine it would be good to change it from `call_format = "nmap -vvv -p {{port}} {{ip}}"` to `call_format = "sudo nmap -vvv -p {{port}} {{ip}} -sU"` but I couldn't figure out how to do this based on the state of my sudp flag


src/scripts/mod.rs
BrendanGlancy, line 1: This is what I mentioned I couldn't figure out

bee-san:
Hey @BrendanGlancy , what references did you use to construct the UDP packet? As you can probably predict I will need to read the docs to review this properly

I am thinking actually take a look at it, merge it, but we don't follow Nmap's way. This will make UDP unreliable, but we'll have it.

Then we can solict people to improve upon it by creating issues for specific problems we experience.

I find it's actually better to merge incomplete features and improve upon them rather than wait an infinite amount of time to have the most perfect feature :)

BrendanGlancy:
@bee-san Thanks for looking at this! If you're talking about `udp_packets` file
[RFC Doc Search](https://datatracker.ietf.org/doc/)


If we go down the list of packets I currently have in there:
1. **SNMP (Simple Network Management Protocol)**:
   - **RFC 1157**: A Simple Network Management Protocol (SNMP).
   - **RFC 1905**: Protocol Operations for SNMPv2.
   - **RFC 2578**: Structure of Management Information Version 2 (SMIv2).
   - **RFC 3416**: Version 2 of the Protocol Operations for the Simple Network Management Protocol (SNMP).

Theres pages for all of those, the one I actually found helpful for creating the packet, was https://datatracker.ietf.org/doc/html/rfc1592#section-3.1.1
Theres tables you'll see if you scroll

2. **DHCP (Dynamic Host Configuration Protocol)**:
   - **RFC 2131**: Dynamic Host Configuration Protocol.
   - **RFC 2132**: DHCP Options and BOOTP Vendor Extensions.

Helpful in creating the packet
https://datatracker.ietf.org/doc/rfc1541/ page 24 table

3. **NTP (Network Time Protocol)**:
   - **RFC 5905**: Network Time Protocol Version 4: Protocol and Algorithms Specification. 

[Stack Overflow source](https://stackoverflow.com/questions/14171366/ntp-request-packet)
image


The one I have isn't even correct it just works in the testing I did

4. **DNS (Domain Name System)**:
   - **RFC 1035**: Domain Names - Implementation and Specification.

Found this on some [medium](https://cabulous.medium.com/dns-message-how-to-read-query-and-response-message-cfebcb4fe817) article, cause the datatracker docs is an insane read for my smoothed over brain
1_heEItB6nil3lq98Cg8EuWw (1)


**TLDR** The packets are a problem, creating them is a pain is was a mix of handwriting and copilot
Testing this stuff is also very tedious as, find an Ip with open UDP port, nmap -sU wait 10-30 minutes compare rustscan.
I will start to work on improving this, as I am still learning about this.

Better news, looking more into nmap they are actually just sending empty packets to most ports on UDP scans, crafting specific payloads for specific well known ports.
I will add that to the TODO list
Also this was helpful for the packets https://github.com/nmap/nmap/blob/master/nmap-service-probes but they have some kind of insane parser and my packets are probably wrong

If you get a chance to look over this, and want me to change other things besides packet generation I will get on that first. 
Otherwise I'll work on improving and adding to the packet crafting, as what I have now is pretty terrible, but it kind of works based on what I have seen.

bee-san:
> find an Ip with open UDP port, 

Hey, so I could actually deploy an AWS EC2 instance and run some stuff on Docker for you to test? Since it's EC2 I believe scannigng is ok, since I (the EC2 owner) will say it's okay 🤷🏻 

> Better news, looking more into nmap they are actually just sending empty packets to most ports on UDP scans, crafting specific payloads for specific well known ports.

This is cool! We have no service probing functionality yet for TCP or UDP :) 

> If you get a chance to look over this, and want me to change other things besides packet generation I will get on that first.

Sure, let me take a quick look :)

src/input.rs
```diff
@@ -150,6 +150,10 @@ pub struct Opts {
     /// A list of comma separated ports to be excluded from scanning. Example: 80,443,8080.
     #[arg(short, long, value_delimiter = ',')]
     pub exclude_ports: Option<Vec<u16>>,
+
+    /// UDP scanning mode, finds udp ports that send back responses
```
bee-san, line 154: s = sudo? or s = scan?

If s = scan, I would remove it :)

src/main.rs
```diff
@@ -83,6 +83,8 @@ fn main() {
     // Added by wasuaje - 01/26/2024:
     // exclude_ports  is an exclusion port list
     //
+    // Added by brendanglancy - 5/19/2024:
+    // sudo is an option to do a udp scan
```
bee-san, line 87:
```suggestion
    // sudp is an option to do a udp scan
```

?

src/scanner/mod.rs
```diff
@@ -130,8 +137,33 @@ impl Scanner {
     ///
     /// Note: `self` must contain `self.ip`.
     async fn scan_socket(&self, socket: SocketAddr) -> io::Result<SocketAddr> {
-        let tries = self.tries.get();
+        if self.sudp {
+            // we try different wait times and packet types
+            // I looked for a crate to create these packets but couldn't find one
```
bee-san, line 142: You are welcome to make a sub-crate in RustScan if you  think a crate is worth it :) 

I think it's https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html , although only if you think it's worth it. It's a lot of overhead to make a sub crate

src/scanner/mod.rs
```diff
@@ -130,8 +137,33 @@ impl Scanner {
...
+            let waits = vec![0, 50, 100, 300];
```
bee-san, line 143: Suggestion: Make the times less even? A firewall would pick up on even timing. Ideally we'd use random times but it's slow and should be in a --quiet option (one for another PR)

Just doing things like 0, 51, 107, 313 etc would be good :)

src/scanner/mod.rs
```diff
@@ -130,8 +137,33 @@ impl Scanner {
...
+            let payloads = vec![
+                craft_dns_query_packet(),
+                craft_ntp_packet(),
+                craft_dhcpc_packet(),
+            ];
 
+            for payload in &payloads {
+                for &wait_ms in &waits {
+                    let wait = Duration::from_millis(wait_ms);
+                    match self.udp_scan(socket, payload.clone(), wait).await {
```
bee-san, line 153: `payload.clone()`

I think you can improve here.

You reference payload &payloads on line 150, and then clone them in a double for loop.

I think this is a bit inefficient, you could instead clone here:

```
for payload in &payloads
```

I am thinking you do not even need to make payloads at all:

```

            let payloads = vec![
                craft_dns_query_packet(),
                craft_ntp_packet(),
                craft_dhcpc_packet(),
            ];
```

I am thinking you can just call the right function in `udp_scan` when it's needed. I have not read all of the code yet though 🙈

bee-san: Ahh ok reading your comment:

> Better news, looking more into nmap they are actually just sending empty packets to most ports on UDP scans, crafting specific payloads for specific well known ports.

Yes, I would look at this a bit more. Crafting 3 packets and hoping the server responds to one is a bit costly. By moving into empty packets you can speed up a lot

src/scanner/mod.rs
```diff
@@ -130,8 +137,33 @@ impl Scanner {
...
+                        Ok(true) => return Ok(socket), // Successful scan
+                        Ok(false) => continue,         // Timed out, try next wait time or payload
+                        Err(e) => return Err(e),       // Some other error occurred
+                    }
+                }
+            }
+            return Err(io::Error::new(
+                io::ErrorKind::TimedOut,
+                "All payloads and wait times failed",
```
bee-san, line 162: I'm not so sure about this error? Is it attached to the If statement? I do not think TimedOut will be good here. And I am not sure we need an error here unless Rust is complaining :)

src/scripts/mod.rs
```diff
@@ -89,6 +89,8 @@ use std::string::ToString;
 use subprocess::{Exec, ExitStatus};
 use text_placeholder::Template;
 
+// Note: I couldn't figure out how to get it to pipe into nmap with the -sU scan
+// since -sU needs sudo
```
bee-san, line 93: I would ignore this 🤷🏻  I am in favour of phasing out Nmap :D

src/udp_packets.rs
```diff
@@ -0,0 +1,293 @@
+pub fn craft_snmptrap_packet() -> Vec<u8> {
```
bee-san, line 1: I would absolutely put the docs you used here from ur comment :D

bee-san (changes requested):
ok heres some lil rust things :) hope that helps!

bee-san:
Thank you SO much!! Such a great PR :) It is crazy you consider yourself a junior developer. I would not even expect this type of code from senior developers 😂

BrendanGlancy:
@bee-san This is like, in the direction of what you were suggesting. 

I'm just not good enough at Rust currently to make payload generation that great. I'm pushing to work on this on a different PC and to "merge incomplete features and improve upon them rather than wait an infinite amount of time to have the most perfect feature " because there's a lot of TODOs that keep popping up with everything I add.

Thank you so much for all the help. It has made this super fun to work on, and I am learning a lot. I'd be interested in the EC2 thing unless it's going to cost you money. 

As always, I'm going to continue to work on this. If you have any more suggestions, I'll work on those. I appreciate the input :)

bee-san:
> I'm just not good enough at Rust currently to make payload generation that great

Nonsense! My first ever Rust program was RustScan, and this code looks fine to me :) 

>  I'd be interested in the EC2 thing unless it's going to cost you money.

Will find time to do that :) 

I have not read your updates yet but I did some reading.

I am thinking version 0.0.1 of this could just be the empty UDP packets like you talked about, and then we can add new packet types for specific things in the future. One per PR would be nice :) 

A common pattern I do is to break the problem down and then build out the absolute minimum viable solution for it (in this case, just empty packets).

But I write the code anticipating the need that there will be more packets in the future. 

However I think here it'll be ok to mostly not do that....

I am not sure how we'd match ports up to specific packets, we'd need a hashmap of port:function() 🤔 

I am thinking:
1. How can I get this merged now?
2. How can we make it easier for people in the future to contribute their own UDP packets? If you make it as easy to contribute to as possible and you make GitHub issues for specific packets I am sure we will get some contributors weighing in :)

BrendanGlancy:
> I am thinking version 0.0.1 of this could just be the empty UDP packets like you talked about, and then we can add new packet types for specific things in the future. One per PR would be nice :)

That is fine, that would just a matter of commenting out some code. When we UDP::bind then => Connect, std::net handles creating the UDP packet headers for us

    pub fn custom_payload(dst_prt: u16) -> Vec<u8>{
        match dst_prt {
            // 53 => craft_dns_query_packet(),
            // 67 => craft_dhcpc_packet(),
            // 123 => craft_ntp_packet(),
            // 161 => craft_snmp_packet(),
            _ => vec![]
        }
    }

> I am not sure how we'd match ports up to specific packets, we'd need a hashmap of port:function() 🤔

> I am thinking:

> 1. How can we get this merged now?

I am not sure besides building what I mention below.

> 2. How can we make it easier for people in the future to contribute their own UDP packets? If you make it as easy to contribute to as possible and you make GitHub issues for specific packets I am sure we will get some contributors weighing in :)

[https://nmap.org/book/nmap-payloads.html] I think having some kind of parser for something like this wouldn't be that difficult. 

The more difficult part is building out some kind of library to make it easier to hard code these payloads, for example theres a bunch of different versions of SNMP payloads they all share the same structure.

IP-packet-encapsulation-of-SNMP-version-2c

I can begin to work on this. Also, I am probably reiterating stuff you already know, for the purpose of making it easier for me to keep track of TODOs on this feature.

bee-san:
want me to review / merge?

BrendanGlancy:
@bee-san yeah sure!

src/input.rs
PsypherPunk, line 154: nitpick: _"…finds UDP ports that…"_

src/scanner/mod.rs
```diff
@@ -191,6 +203,90 @@ impl Scanner {
...
+    /// # Example
+    ///
+    /// ```compile_fail
+    /// # use std::net::{IpAddr, Ipv6Addr, SocketAddr};
+    /// let port: u16 = 80;
+    /// // ip is an IpAddr type
+    /// let ip = IpAddr::V6(Ipv6Addr::new(0, 0, 0, 0, 0, 0, 0, 1));
+    /// let socket = SocketAddr::new(ip, port);
+    /// scanner.udp_bind(socket);
+    /// // returns Result which is either Ok(stream) for port is open, or Er for port is closed.
```
PsypherPunk, line 217: nitpick: _"…or Err for port…"_.

src/scanner/mod.rs
```diff
@@ -191,6 +203,90 @@ impl Scanner {
...
+    /// // returns Result which is either Ok(true) if response received, or Ok(false) if timed out.
+    /// // Err is returned for other I/O errors.
+    async fn udp_scan(
+        &self,
+        socket: SocketAddr,
+        payload: &[u8],
+        wait: Duration,
+    ) -> io::Result<bool> {
+        match self.udp_bind(socket).await {
+            Ok(x) => {
```
PsypherPunk, line 252: suggestion: `x` here is a `UdpSocket` (I think); `socket` might be more descriptive than a single-character variable name.

src/scanner/mod.rs
```diff
@@ -191,6 +203,90 @@ impl Scanner {
...
+                let mut buf = [0u8; 1024];
+
+                x.connect(socket).await?;
+                x.send(&payload).await?;
+
+                match io::timeout(wait, x.recv(&mut buf)).await {
+                    Ok(size) => {
+                        println!("Recived {} bytes", size);
```
PsypherPunk, line 260: suggestion: use the `debug!` macro instead of `println!` throughout.

Probably better to use the already-imported `debug!` macro (or another level, if appropriate), rather than `println!` throughout, so it can filtered.

src/udp_packets.rs
```diff
@@ -0,0 +1,277 @@
+pub mod udp_payload {
+    // const MAX_PAYLOAD_IPV4: usize = 65507; // 65,535 - 20 (IP header) - 8 (UDP header)
+    // const MAX_PAYLOAD_IPV6: usize = 65527; // 65,535 - 40 (IPv6 header) - 8 (UDP header)
+
+    /// From RFC 768
```
PsypherPunk, line 5: question: worth a link to the actual RFC?

i.e. `/// From [RFC 768](https://www.rfc-editor.org/rfc/rfc768.txt)`

src/scanner/mod.rs
```diff
@@ -130,8 +134,22 @@ impl Scanner {
     ///
     /// Note: `self` must contain `self.ip`.
     async fn scan_socket(&self, socket: SocketAddr) -> io::Result<SocketAddr> {
-        let tries = self.tries.get();
+        if self.udp {
+            let waits = vec![0, 51, 107, 313];
```
PsypherPunk, line 138: question: can I check where these values have come from?

Are they particularly meaningful? And worth documenting?

src/udp_packets.rs
```diff
@@ -0,0 +1,277 @@
...
+        let pdu_type = vec![0xa4, 0x82]; // PDU type for SNMP Trap
+        let enterprise_oid = vec![0x06, 0x09, 0x2b, 0x06, 0x01, 0x02, 0x01, 0x01, 0x01, 0x00]; // Enterprise OID
+
+        // Combine the components into a single SNMP Trap packet
+        [version, community_string, pdu_type, enterprise_oid].concat()
+    }
+
+    /// Crafts an SNMP packet based on the structure from RFC 1592.
+    ///
+    /// RFC 1592 defines the structure of an SNMP packet as follows
```
PsypherPunk, line 53: question: worth linking to the actual RFC?

As above, is it worth referencing the official doc?

src/udp_packets.rs
```diff
@@ -0,0 +1,277 @@
...
+        // Query Type: A (host address)
+        packet.extend(&[0x00, 0x01]);
+
+        // Query Class: IN (internet)
+        packet.extend(&[0x00, 0x01]);
+
+        packet
+    }
+
+    /// Creates the simplest ntp packet possibl
```
PsypherPunk, line 265: nitpick: casing/typo.

Should be `/// Creates the simplest NTP packet possible`…?

src/tui.rs
```diff
@@ -93,6 +93,7 @@ macro_rules! funny_opening {
             "I scanned ports so fast, even my computer was surprised.",
             "RustScan: Where '404 Not Found' meets '200 OK'.",
             "RustScan: Exploring the digital landscape, one IP at a time.",
+            "DeathStarMafia was here 🚀",
```
PsypherPunk, line 96: question: is this suitable?

One for @bee-san: is this okay to add? Not sure there are other, named individuals/orgs. in this list…?

PsypherPunk (commented):
Couple of minor quibbles with typos. and a couple of questions but this is very, very nice.

src/scanner/mod.rs
PsypherPunk: Additionally, typo. in `Recived`.

src/tui.rs
BrendanGlancy: Lol that's my CTF team (I did not name it), but I'll change this. But yes the 0day guy added himself to this but maybe he's an exception as he's tiktok famous

src/udp_packets.rs
BrendanGlancy: These links were a suggestion from @bee-san, which I may have misinterpreted. For now, this can all be removed, as we're sending blank packets anyway (from what I understand)

PsypherPunk: Oh, I'm not suggesting removing these docs. (they're a great addition) just citing the original for further reading.

src/tui.rs
PsypherPunk: I didn't realise "0day" was a person; thought it was a reference to the _thing_. Shows what I know.

Kudos on the CTF team name though.

src/scanner/mod.rs
BrendanGlancy: No, they were 50, 100, 300.

bee-san suggested 
> Suggestion: Make the times less even? A firewall would pick up on even timing. Ideally we'd use random times 
> but it's slow and should be in a --quiet option (one for another PR)
> Just doing things like 0, 51, 107, 313 etc would be good :)

Would've never thought of this myself

src/scanner/mod.rs
BrendanGlancy: I agree, we cannot name it socket but I've changed it to udp_socket

there is also another section of the code (I didn't write):
```rust
// Location scanner/mod.rs - scan_socket 155
        for nr_try in 1..=tries {
            match self.connect(socket).await {
                Ok(x) => {
                    debug!(
                        "Connection was successful, shutting down stream {}",
                        &socket
                    );
                    if let Err(e) = x.shutdown(Shutdown::Both) {
                        debug!("Shutdown stream error {}", &e);
                    }
                    self.fmt_ports(socket);

                    debug!("Return Ok after {} tries", nr_try);
                    return Ok(socket);
                }
```
Very similar naming convention, I can rename this to tcp_socket while I'm in here.

src/tui.rs
bee-san: @PsypherPunk 
@BrendanGlancy yes, 0day is one of my friends. his dog passed away sadly and i wanted to commemorate him (Ollie), 0day put in a PR to add his own name so I thought it'd be ok 😅 

@BrendanGlancy I will let you add your name here if you add 3 more networking related / hacking jokes 😉

src/scanner/mod.rs
bee-san: @PsypherPunk  Because of the way UDP works we don't know how long to wait to assume failure. Normally TCP would always get back to us, and if it doesn't we wait 1 second I think.

We need to tell UDP how long to wait for before moving on, I was thinking if we send multiple UDP packets at even timestamps the firewall would easily pick up on it. Uneven doesn't help _so_ much but it's slightly better than nothing.

Something I realised is we have `0` here. Reading the code:

https://github.com/RustScan/RustScan/blob/52bfcf048db5e10bcdd887abb350c85606331760/src/scanner/mod.rs#L258

Does this imply we wait `0` seconds before assuming it's closed?
And thinking more about it do we need a vector of timeouts here at all? 

The normal TCP scan waits 1.5 seconds:
https://github.com/RustScan/RustScan/blob/52bfcf048db5e10bcdd887abb350c85606331760/src/input.rs#L116

I think you should actually parse the `timeout` input down here and use that instead? That way in the future we can update this to a random timeout if we want to :)

Sorry, only just caught this 🙈

src/scanner/mod.rs
PsypherPunk: @BrendanGlancy, I can see this has been updated—thanks—but there's still a typo. in `Recieved`, I'm afraid.

src/scanner/mod.rs
bee-san: Yes actually thinking about this more, there's other functions which use the timeout given at the argparsing level so I think we should not include this vector and just parse the timeout from the argparsing part (like the other tcp scanner does)

and I think a nice lil feature would be some code that can alter that timeout more in depth, like randomised timeouts

bee-san:
@BrendanGlancy  are you in my discord? http://discord.skerritt.blog 

got an ec2 instance you can scan for udp but dont want to publish the IP to the public haha

PsypherPunk:
> got an ec2 instance you can scan for udp but dont want to publish the IP to the public haha

FWIW, I've been spinning up a quick UDP socket via Python to test this (or at least making sure it picks up the right port):

```python
import socket


UDP_IP = "127.0.0.1"
UDP_PORT = 5005

sock = socket.socket(
    socket.AF_INET, # Internet
    socket.SOCK_DGRAM,
)
sock.bind((UDP_IP, UDP_PORT))

while True:
    data, addr = sock.recvfrom(1024)
    print(f"Received message: {data}")
```

BrendanGlancy:
@bee-san Yes I just joined, but I already got in trouble with the mod. hopefully, I'll have a commit with the requested changes tonight

@PsypherPunk Sorry about all the typos, I've never had a real software engineering job where someone reviews my code I didn't realize I was that sloppy

Really appreciate all the feedback from the both of you :)

PsypherPunk:
> Ahh it's okay! It's good code :) I really think that after you update it to use the timeout from the args we can get this merged, I see no other issues @PsypherPunk ?

One outstanding typo. in `Recieved` and I'm done 😁

BrendanGlancy:
@bee-san 

Hopefully this is what you meant
```rust
    async fn scan_socket(&self, socket: SocketAddr) -> io::Result<SocketAddr> {
        if self.udp {
            let payload = cust_payload(socket.port());

            let tries = self.tries.get();
            for _ in 1..=tries {
                match self.udp_scan(socket, &payload, self.timeout).await {
                    Ok(true) => return Ok(socket),
                    Ok(false) => continue,
                    Err(e) => return Err(e),
                }
            }
            return Ok(socket);
        }
```

bee-san:
> @bee-san
> 
> Hopefully this is what you meant
> 
> ```rust
>     async fn scan_socket(&self, socket: SocketAddr) -> io::Result<SocketAddr> {
>         if self.udp {
>             let payload = cust_payload(socket.port());
> 
>             let tries = self.tries.get();
>             for _ in 1..=tries {
>                 match self.udp_scan(socket, &payload, self.timeout).await {
>                     Ok(true) => return Ok(socket),
>                     Ok(false) => continue,
>                     Err(e) => return Err(e),
>                 }
>             }
>             return Ok(socket);
>         }
> ```

Amazing!!

1 small thing, you have this:

```
tries
```

This looks like a user defined argument, but:
1. I can't see any default for it
2. I can't see it in the arguments

And I think it'd be better named something like `udp_retries`, to indicate its for retries specifically for udp :)

If I'm wrong let me know, it's early for me but this looks good!

src/scanner/mod.rs
```diff
@@ -191,6 +202,90 @@ impl Scanner {
...
+    /// Formats and prints the port status
+    fn fmt_ports(&self, socket: SocketAddr) {
+        if !self.greppable {
+            if self.accessible {
+                println!("Open {socket}");
+            } else {
+                println!("Open {}", socket.to_string().purple());
+            }
+        }
+    }
```
bee-san, line 288: Ticket to make in the future, we need to move all of this printing stuff into its own macro in the TUI file :)

bee-san:
Ok, CI is failing the lint check and I'm wrong, we have tries here:
https://github.com/RustScan/RustScan/blob/cec858fc2d80dec786e9f1e3f95d740cdb018bc4/src/input.rs#L122

bee-san:
ok for me just CI needs to pass and we can merge :)

BrendanGlancy:
> > @bee-san
> > Hopefully this is what you meant
> > ```rust
> >     async fn scan_socket(&self, socket: SocketAddr) -> io::Result<SocketAddr> {
> >         if self.udp {
> >             let payload = cust_payload(socket.port());
> > 
> >             let tries = self.tries.get();
> >             for _ in 1..=tries {
> >                 match self.udp_scan(socket, &payload, self.timeout).await {
> >                     Ok(true) => return Ok(socket),
> >                     Ok(false) => continue,
> >                     Err(e) => return Err(e),
> >                 }
> >             }
> >             return Ok(socket);
> >         }
> > ```
> 
> Amazing!!
> 
> 1 small thing, you have this:
> 
> ```
> tries
> ```
> 
> This looks like a user defined argument, but:
> 
> 1. I can't see any default for it
> 2. I can't see it in the arguments
> 
> And I think it'd be better named something like `udp_retries`, to indicate its for retries specifically for udp :)
> 
> If I'm wrong let me know, it's early for me but this looks good!

I'll change it, I copied it from the tcp scanner

bee-san (approved):

bee-san:
Wow! Crazy! UDP!!! 🥳 Merged!!!!

BrendanGlancy:
You guys have been awesome to work with.

bee-san:
@BrendanGlancy  feel free to make more PRs, this was fantastic 🥳

PR #598, opened 2024-05-30
ci: refactor ci/cd and extends matrix build

LeoFVO:
Hey @bee-san !

I'm wanted to improve the delivery and ease of use for people.
I improve the release CI/CD and publish more format of RustScan.

Now, when you do a release, the release will be delivered with:
- rustscan.exe
- ruscan (linux x86_64 binary)
- debian package
- cross-build as .zip
- cross-build as .tar.xz

I fixed a minor bug in the CI.
And I tried to add support for aarch64-linux build, but it's too complicated for the moment...

Feel free to ask if you need more information !

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ repository = "https://github.com/rustscan/rustscan"
 license = "MIT"
 keywords = ["port", "scanning", "nmap"]
 categories = ["command-line-utilities"]
-readme="README.md"
+readme = "README.md"
 exclude = [
     ".github/*",
     "pictures/*",
@@ -24,6 +24,7 @@ futures = "0.3"
 rlimit = "0.10.1"
 log = "0.4.0"
 env_logger = "0.11.3"
+anstream = "=0.6.14"
 dirs = "5.0.1"
 gcd = "2.0.1"
 rand = "0.8.5"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,13 @@
 name: Build
 on:
   pull_request:
+    paths-ignore:
+      - '.github/**'
   push:
     branches:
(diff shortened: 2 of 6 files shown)


LeoFVO:
The pipeline is failing on the job trying to upload assets to the release, so it's “normal” that's failing in the context of a PR, but this work well in real case ;)

LeoFVO:
> The pipeline is failing on the job trying to upload assets to the release, so it's “normal” that's failing in the context of a PR, but this work well in real case ;)

Fixed this minor issues :)

Cargo.toml
```diff
@@ -24,6 +24,7 @@ futures = "0.3"
 rlimit = "0.10.1"
 log = "0.4.0"
 env_logger = "0.11.3"
+anstream = "=0.6.14"
```
bee-san, line 27: 👀  what's this for?

bee-san:
looks very good, thank you so much!!!

Cargo.toml
LeoFVO: Got this error when running tests:
```bash
cargo test 
error: failed to select a version for `env_logger`.
    ... required by package `rustscan v2.2.4-rc1 (/Users/leofvo/Developer/opensource/RustScan)`
versions that meet the requirements `^0.11.3` (locked to 0.11.3) are: 0.11.3

the package `rustscan` depends on `env_logger`, with features: `anstream` but `env_logger` does not have these features.
 It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.


failed to select a version for `env_logger` which could resolve this conflict
```

So the way I fixed was by adding the dependency explicitly here :/

bee-san:
ok thank youuu!

PR #610, opened 2024-07-05
Removed unnecessary `.clone()`s

Arvmor:
### Explanation
the `.clone()`s displayed below are replaced with `.as_ref()` for enhance performance!

https://github.com/RustScan/RustScan/blob/3873117324a1af07c96d94dca59827665f13f1d0/src/input.rs#L211-L218

## Fix
https://github.com/RustScan/RustScan/blob/ee7d00ec84bc1e8ae6b1d67e90119150ed2dff56/src/input.rs#L211-L218

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -210,8 +210,8 @@ impl Opts {
 
         // Only use top ports when the user asks for them
         if self.top && config.ports.is_some() {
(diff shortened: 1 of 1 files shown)


Arvmor:
> ayy i love PRs like this! Thank you so much!!

Bhaha thanks, my cat had a surgery, had to distract myself with something

PR #616, opened 2024-07-09
Match crate license to source

### LINKED ISSUE #615: License discrepancy between crate and source
Thanks for the version bump and crates revival -- makes packaging for Fedora go much more smoothly

I noticed however that the crate and source disagree on which license is being applied. The crate uses MIT while the source uses GPL-3.0. Would we be able to match these licenses to avoid issues downstream?

ppfeister:
Fixes #615 

Is it possible to republish crates with a new build number (-1 -2) rather than a bump in patch number (x.x.1 x.x.2) that triggers notifications? Or overwrite prior builds? I know you can't on PyPI but haven't tried on crates[.]io. If this doesn't make its way into the crate until the next release I'll just continue to build from source for now, using my existing spec file.


bee-san:
Hiii! Thanks for this!! Since we don't have a CLA that means the code is licensed under what people thought the license was when contributing.

Sooo I think I will put a post in my Discord letting contributors know and seeing if they have any issues with this, in case they thought it was MIT when writing code :)

bee-san (approved):
Okay, couldnt find anyone that disapproved :)

PR #618, opened 2024-07-11
Using references in `init_scripts()` to avoid unnecessary `.clone()`s

Arvmor:
### ISSUE
the `init_scripts()` function includes a few unnecessary `.clone()`s which could be easily avoided through using references, the error handling is slightly verbose and if statements could also be less verbose by using `if let`.

https://github.com/RustScan/RustScan/blob/9f35cd43bc4124c479d54a0664503b6165fe9dfd/src/scripts/mod.rs#L129-L148

### FIX
https://github.com/RustScan/RustScan/blob/033e2810f28be22610030e1974bf7fb351c9c4a6/src/scripts/mod.rs#L120-L135

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -45,7 +45,7 @@ fn main() {
 
     debug!("Main() `opts` arguments are {:?}", opts);
 
-    let scripts_to_run: Vec<ScriptFile> = match init_scripts(opts.scripts) {
+    let scripts_to_run: Vec<ScriptFile> = match init_scripts(&opts.scripts) {
         Ok(scripts_to_run) => scripts_to_run,
         Err(e) => {
             warning!(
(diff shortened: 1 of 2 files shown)


Arvmor:
### ISSUE
https://github.com/RustScan/RustScan/blob/9f35cd43bc4124c479d54a0664503b6165fe9dfd/src/scripts/mod.rs#L129-L148

### FIX
https://github.com/RustScan/RustScan/blob/033e2810f28be22610030e1974bf7fb351c9c4a6/src/scripts/mod.rs#L120-L135

PsypherPunk (approved):
Nice; I am entirely in favour of this.

PsypherPunk:
@Arvmor, just sorting some failures in the build pipeline but this should be good to go.

Arvmor:
> @Arvmor, just sorting some failures in the build pipeline but this should be good to go.

Great!! If anything comes up please let me know.

PR #620, opened 2024-07-15
BugFix: Allow IPV6 addresses to be parsed into NMAP

israelrg:
Hi,
My ISP defaults to IPV6. I found that the IPV6 version of the domain www.google.es can't be scanned. I attached you a prof one with the domain name translated into IPV6 and the other with the IPV4 address that nslookup resolves.

Captura de pantalla 2024-07-15 232549

This is what I'm referring to

`Open [2a00:1450:4003:80f::2003]:443
[~] Starting Script(s)
2a00:1450:4003:80f::2003 looks like an IPv6 target specification -- you have to use the -6 option.`

I added support to IPV6 by providing the version, both 4 & 6

Regards 🙂

### DIFF
--- a/src/scripts/mod.rs
+++ b/src/scripts/mod.rs
@@ -76,7 +76,6 @@
 #![allow(clippy::module_name_repetitions)]
 
 use crate::input::ScriptsRequired;
-use anstream::println;
 use anyhow::{anyhow, Result};
 use log::debug;
 use serde_derive::{Deserialize, Serialize};
--- a/src/scripts/mod.rs
+++ b/src/scripts/mod.rs
@@ -88,11 +88,12 @@ use std::path::PathBuf;
 use std::string::ToString;
 use subprocess::{Exec, ExitStatus};
 use text_placeholder::Template;
+use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


israelrg:
Here is the prof that the fix works by providing the IP version

Captura de pantalla 2024-07-15 233452

PsypherPunk:
@israelrg, this is nice. In principle I agree with the change but instead of relying on a failed `from_str()` call in the `match` statement, can we make better use of the `IpAddr` we already store in the `Script`?

At the moment it looks like the `ExecPartsScript`/`ExecParts` just calls `to_string()` on the `ip` in the `Script`. However, at that point we should be able to determine which variant it is…?

Also, probably worth adding a test or two in this module to validate the behaviour.

israelrg:
Hi @PsypherPunk,
Can you verify if it's better now?
Regards

PsypherPunk:
@israelrg, yep: that looks good.

The only outstanding comment would be to extend the tests to validate the change.

PsypherPunk (approved):
I'm happy to accept this: although I think adding tests would be useful, I don't think our code is currently structured in a way to easily do that (other than manually, which I have).

PR #636, opened 2024-07-30
docs: fix homebrew formula link

chenrui333:
right now the link is for rust crate, update to use the brew formula page.


bee-san (approved):
thanbks!!

PR #643, opened 2024-08-11
Statically generated payloads 

BrendanGlancy:
This is a work in progress; I don't know if all the packets are being generated correctly, as it is hard to test.

I am looking to get more eyes on this, as I have already made several decisions myself that other more experienced contributors could've made better calls on.

Basically, we are creating a BTreeMap of [Ports[], Payloads[]], which works better than anything else I have put together so far in terms of UDP scanning accuracy. 

For example, port 67 and 68 are common DHCP ports so sending them the same packet to test for a response is easy with the format:
```
[[67,68],[payload]]
```

I am fairly certain I am not violating the Nmap license by using this payload database, as Rustscan is also free software, and this database is deprecated.

The current, more complete Nmap service probing database is incredibly difficult to parse, and incorporating something like that into Rustscan would cause the project to become rigid. This older version allows easy contribution in terms of adding UDP payloads and could be fairly easy to adapt to something like Yaml.


BrendanGlancy:
@PsypherPunk not sure how to request review besides tag

bee-san:
> @BrendanGlancy, I'm away for a few days but will take a look ASAP. Looks like the build is failing at the moment though so probably worth determining where that's coming from.
> 
> On a less fun note, we might need to clarify the position around reusing Nmap's code. Although it's open-source, it uses its own ["Nmap Public Source License"](https://svn.nmap.org/nmap/LICENSE) which has this explicit clause:
> 
> > These additional terms mean that You may not distribute Covered Software or Derivative Works under plain GPL terms without special permission from Licensor.
> 
> …and as RustScan is GPL, that might rule this out.

That license specifically defines "GPL" as GPL version 2

> * "GPL" means the GNU General Public License Version 2, as published
  by the Free Software Foundation and provided in Exhibit A.

Under "Definitions"

RustScan uses GPL 3.0 https://github.com/RustScan/RustScan?tab=GPL-3.0-1-ov-file#readme

Also, their document here mentions
https://nmap.org/npsl/ 

That it's "open sourcd" and links to this definition:
https://opensource.org/osd

Which says:

> The license must allow modifications and derived works, and must allow them to be distributed under the same terms as the license of the original software.

I do not see any issues here, so long as we do not commercialise RustScan with private code it will be ok :)

build.rs
```diff
@@ -0,0 +1,159 @@
+use std::collections::BTreeMap;
```
bee-san, line 1: 1. We should have comments on what this file does :)
2. I am not a super big fan of using Rust to build files :( We should use something like Bash or Python I think, as it's easier to read & contribute to?

src/generated.rs
```diff
@@ -0,0 +1,59 @@
...
+    map.insert(vec![11211], vec![0,16,0,1,0]);
+    map.insert(vec![17185], vec![114,55,114,55,0,0,0,0,0,0,0,2,85,85,85,85,0,0,0,1,0,0,0,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,255,255,85,16,0,0,0,60,0,0,0,3,0,0,0,2,0,0,0,0,0,0,0,0]);
+    map.insert(vec![26000,26001,26002,26003,27960,27961,27962,27963,30720,30721,30722,30723,44400], vec![255,255,255,255,234]);
+    map.insert(vec![27015,27016,27017,27018,27019,27020,27021,27022,27023,27024,27025,27026,27027,27028,27029], vec![255,255,255,255,206,238,224]);
+    map.insert(vec![27444], vec![68,173]);
+    map.insert(vec![27910,27911,27912,27913], vec![255,255,255,255]);
+    map.insert(vec![31337], vec![206,99,209,210,22,231,19,207,56,165,165,134,178,117,75,153,170,50,88]);
+    map.insert(vec![34555], vec![]);
+    map.insert(vec![64738], vec![0,0,0,0,171,205,239]);
+    map
```
bee-san, line 58: this is a big file, do we have to make this at run time? can we not build this before hand and consume it with yaml or something? or even have a perfect hash map https://skerritt.blog/perfect-hash-tables/ ?

bee-san (commented):
just 2 initial comments :D

src/generated.rs
BrendanGlancy: > I am not a super big fan of using Rust to build files :( We should use something like Bash or Python I think, as it's easier to read & contribute to?

Do you mean rewriting the entire file parser in Python?

I not entirely sure how this could be done, maybe you know @HDTHREE

> this is a big file, do we have to make this at run time? can we not build this before hand and consume it with yaml or something? or even have a perfect hash map https://skerritt.blog/perfect-hash-tables/ ?

A couple of things could happen with this file; we could .gitignore it. I cannot get any version of the Rust LSP to format it for me, so I think it breaks part of the CI. I was super weirded out by the whole code generation at compile time thing, but yes the way we are actually consuming this map at runtime is something I can work on. 

> @BrendanGlancy, I'm away for a few days but will take a look ASAP. Looks like the build is failing at the moment though so probably worth determining where that's coming from.

@PsypherPunk && Bee no rush on any of this, just wanted to get your thoughts on this before I moved any further forward as this PR is already the size of the Epstein list.

bee-san: > Do you mean rewriting the entire file parser in Python?

I am a bit sick so I didn't take a proper look, but I think your `build.rs` parses the nmap file and turns it into another Rust file to use, right?

typically for scripts like that I see them as Python / Bash and they go into a seperate `/scripts` folder or something or maybe `fixtures`  🤔 

> A couple of things could happen with this file; we could .gitignore it. I cannot get any version of the Rust LSP to format it for me, so I think it breaks part of the CI. I was super weirded out by the whole code generation at compile time thing, but yes the way we are actually consuming and creating this map at runtime is something I can work on.

yeah that weirds me out too, i think generally the PR looks fine though :)

> @PsypherPunk && Bee no rush on any of this, just wanted to get your thoughts on this before I moved any further forward as this PR is already the size of the Epstein list.

PR is failing partially because cargo format cant format `build,rs`, another win for not putting it into rust ;)

i think trying one of these:
https://www.reddit.com/r/rust/comments/1c6ceuu/how_to_force_cargo_fmt_to_ignore_some_files/

and then running cargo format locally to see if it works or not.

other issue is casued by https://github.com/RustScan/RustScan/issues/527

this has a fix:

The temporary solution is
change
let arguments = shell_words::split(&to_run).expect("Failed to parse script arguments");
to
let arguments:Vec<String> =to_run.split_whitespace().map(|s| s.to_string()).collect();

does that work? 🤔

BrendanGlancy: > I am a bit sick so I didn't take a proper look, but I think your build.rs parses the nmap file and turns it into another Rust file to use, right?

> typically for scripts like that I see them as Python / Bash and they go into a seperate /scripts folder or something or maybe fixtures 🤔

The problem with using python for this is, and I am not 100% sure about this, we will not be able to put the values for the BTreeMap (or whatever datastructure) during compile time.

bee-san:
Concerning it's breaking on Windows... I was hoping it would be temporary.......

image

Does this mean it fails when trying to compile futures-lite? 

once_cell does appear to compile for windows :(

BrendanGlancy:
Oh, sorry, I got distracted. I made the Python version of the parser over the weekend. (https://github.com/BrendanGlancy/RustScan/blob/dev/services.py) It kind of makes everything even more complicated

I don't have a Windows box; I'll have to spin up a VM and look into this. But...

Ideally, this was just supposed to work, and we could move on to TCP service probing; I should've turned on the CI stuff on my branch before I submitted this. 

I think the solution is to hand roll our own service probing 'database.' Should we use YAML?

BrendanGlancy:
@bee-san  what is up with the Linux CI test? I'm not even sure what could be causing that

bee-san:
> @bee-san what is up with the Linux CI test? I'm not even sure what could be causing that

semver violation, let me re-run it. in semantic versioning something like "1.0.0dev" is not allowed. I can't find proof  of this anywhere, so I am rerunning in case it just broke :)

bee-san:
ok this is weird... i am not sure...... :(

PsypherPunk:
In the `env` for the failing step you can see `TAG: dev`. In the previous step that `if` clause must not be firing so the `name` is staying at its default of `dev`.

Looks like it's passing on `macos` as that step would appear to miss `macos` entirely (!?); passing on `windows` as it doesn't make use of `TAG`.

Oh, hang on: should this even be running? Do we actually want to make a release based on pull-request? Build/test definitely, but actually release…? The above is likely to be because it was a pull-request that triggered it, not a tag (therefore there's no Git tag from which to derive `TAG`).

bee-san:
> In the `env` for the failing step you can see `TAG: dev`. In the previous step that `if` clause must not be firing so the `name` is staying at its default of `dev`.
> 
> Looks like it's passing on `macos` as that step would appear to miss `macos` entirely (!?); passing on `windows` as it doesn't make use of `TAG`.
> 
> Oh, hang on: should this even be running? Do we actually want to make a release based on pull-request? Build/test definitely, but actually release…? The above is likely to be because it was a pull-request that triggered it, not a tag (therefore there's no Git tag from which to derive `TAG`).

I think @CMNatic wanted to do PR-based releases for the docker image? We can disable that CI though I don't mind so long as no one else does!!

Thank you for your investigation

BrendanGlancy:
@bee-san So I think at this point everything is fixed except the dev tag in the CI thing

bee-san:
@BrendanGlancy talked to some people, we think its cause its in a PR it doesn't like it but hopefully on release it will like it haha

bee-san:
merged!!

PR #644, opened 2024-08-12
chore: fix some comments

drawdrop:


BrendanGlancy (approved):
lgtm

src/udp_packets.rs
```diff
@@ -14,7 +14,7 @@ pub mod udp_payload {
     /// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
     /// |            Length             |           Checksum            |
     /// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
-    /// We append the data, based on the port the the header
+    /// We append the data, based on the port the header
```
PsypherPunk, line 17: Although I agree with the change, does this sentence make sense? 

> _"…based on the port the header"_

PsypherPunk: Ah, should this be "the port **in** the header"?

BrendanGlancy: "We append the data, based on the port"

PsypherPunk: As in that should be the complete sentence?

I think that makes it a bit ambiguous: it's below a diagram with both "Source Port" and "Destination Port"—to which one is it referring? Or neither?

I think it'd be useful to clarify that as part of this, if possible…?

BrendanGlancy: Ahh then it should be "We append the data, based on the destination port"

PsypherPunk (approved):
Merging, even with the outstanding comment; will raise a PR to fix that as it's technically separate to the changes herein.

PsypherPunk (changes requested):
Scratch that: I didn't see the merge conflict.

@drawdrop, if you can fix the merge conflict, this should be good to merge.

bee-san:
Fixed conflits myself :)

PR #658, opened 2024-09-13
Fix cargo clippy warnings

bee-san:


BrendanGlancy (commented):
don't know what this is lgtm

src/input.rs
```diff
@@ -271,7 +271,7 @@ pub struct Config {
 
 #[cfg(not(tarpaulin_include))]
 #[allow(clippy::doc_link_with_quotes)]
-#[allow(clippy::manual_unwrap_or_default)]
+#[allow(clippy::unwrap_or_default)]
 impl Config {
     /// Reads the configuration file with TOML format and parses it into a
     /// Config struct.
```
PsypherPunk, line 277: I can't comment on the actual in question but `clippy` complains similarly for me locally; replacing the current assignment of `content` with the following resolves the complaint, tests pass, etc:

```rust
if config_path.exists() {
    content = fs::read_to_string(config_path).unwrap_or_default();
}
```

PR #591, opened 2024-05-15
refactor: remove structopt in favour of clap

PsypherPunk:
Prompted by the `atty` issue raised by Dependabot, move to `clap` as this doesn't have that dependency and, as per the [`structopt` README](https://github.com/TeXitoi/structopt#maintenance), that crate is now in maintenance mode.

**Note** that they aren't _exactly_ the same:

Previous:

```sh
rustscan 2.2.3
Fast Port Scanner built in Rust. WARNING Do not use this program against sensitive infrastructure since the specified
server may not be able to handle this many socket connections at once. - Discord  <http://discord.skerritt.blog> -
GitHub <https://github.com/RustScan/RustScan>

USAGE:
    rustscan [FLAGS] [OPTIONS] [-- <command>...]

FLAGS:
        --accessible    Accessible mode. Turns off features which negatively affect screen readers
    -g, --greppable     Greppable mode. Only output the ports. No Nmap. Useful for grep or outputting to a file
    -h, --help          Prints help information
    -n, --no-config     Whether to ignore the configuration file or not
        --top           Use the top 1000 ports
    -V, --version       Prints version information

OPTIONS:
    -a, --addresses <addresses>...            A comma-delimited list or newline-delimited file of separated CIDRs, IPs,
                                              or hosts to be scanned
    -b, --batch-size <batch-size>             The batch size for port scanning, it increases or slows the speed of
                                              scanning. Depends on the open file limit of your OS.  If you do 65535 it
                                              will do every port at the same time. Although, your OS may not support
                                              this [default: 4500]
    -c, --config-path <config-path>           Custom path to config file
    -e, --exclude-ports <exclude-ports>...    A list of comma separated ports to be excluded from scanning. Example:
                                              80,443,8080
    -p, --ports <ports>...                    A list of comma separated ports to be scanned. Example: 80,443,8080
    -r, --range <range>                       A range of ports with format start-end. Example: 1-1000
        --resolver <resolver>                 A comma-delimited list or file of DNS resolvers
        --scan-order <scan-order>             The order of scanning to be performed. The "serial" option will scan ports
                                              in ascending order while the "random" option will scan ports randomly
                                              [default: serial]  [possible values: Serial, Random]
        --scripts <scripts>                   Level of scripting required for the run [default: default]  [possible
                                              values: None, Default, Custom]
    -t, --timeout <timeout>                   The timeout in milliseconds before a port is assumed to be closed
                                              [default: 1500]
[...]

Issue #424, opened 2021-12-28
[Feature Request] make the range -r  do  ----->  1-79, 81-442,444-8079,8081-65535


gprime31:
[Feature Request] make the range -r  do  ----->  1-79, 81-442,444-8079,8081-65535  exclude 80,443,8080 but scan the rest.
so I can do multiple ranges, or make a flag to exclude unwanted ports in a range. whatever is easier.
unless there is another way to do this I might have missed.
Thanks.
Great tool!!

Autumn-27:
May I ask if the range parameter can support 1-6, 11200-300

PR #661, opened 2024-09-24
support --range 1-22,661-669,8888

Autumn-27:
Supports port range optional range and single port.
example:rustscan -a 127.0.0.1  --range 1-22,661-669,8888


PsypherPunk:
Thanks for the PR. It'd be useful to understand what issues you were running into that prompted this change…?

Currently, there's a specific message around not having ranges and individual ports used; it'd be useful to understand if there's a reason for that or if it's just historical:

```sh
the argument '--ports <PORTS>' cannot be used with '--range <RANGE>'
```

That said, if `--range` starts managing both single ports and ranges, those parameters could be confusing and if folks are on board with this change it might be worth deprecating one or the other. I'll wait for other views on that.

Also, I'd suggest any comments are in English to be consistent with the existing codebase.

Autumn-27:
https://github.com/RustScan/RustScan/issues/424
I have also seen someone mention it in the issue before, because currently I am doing integrated scanning, such as scanning the top 100 ports using 1-20, 808080-8088, etc. If it cannot be supported, I need to split all ports into individual ports - ports 1, 2, 3, 4... 20, 8080808081... 8088. There is no problem with a small number of ports, but if there are too many ports, it will exceed the command length limit at runtime. Hope it can be supported

Autumn-27:
I think managing ports through just one is enough~

Autumn-27:
I think this is more convenient

bee-san:
> reason for that or if it's just historical

This took me back! This was a big, big argument we had. I think if I recall 4 years ago it simply was not possible for us to have both arguments at the same time. This is because the argument parsing library we used at the time was not mature enough to support these two things. It sounds insane now, but Rust was a different beast 4 years ago 😂  (I can't find our discussion on GitHub, might be in the old RustScan Discord)

> That said, if --range starts managing both single ports and ranges, those parameters could be confusing and if folks are on board with this change it might be worth deprecating one or the other. I'll wait for other views on that.

Yes I agree. There's also the speed issue, range now needs the ability to parse single ports which adds another loop into our code. For arguments it's fine, but I wonder if `parse_range` is used else where for example in the TOML definition of ports, it may slow down quite a bit.

I think a good solution is to keep both, and to allow both to be used at the same time. 

We use Clap now, so I would hope it's possible to use both arguments at the same time :) https://github.com/RustScan/RustScan/pull/591 

Also, the CI is failing. I'd suggest running these locally:
* Cargo test
* Cargo clippy
* Cargo fmt

And if all are okay, the CI will generally be ok :)

Autumn-27:
I've considered supporting both parameters, but this still requires splitting the top 1000 common ports and then splicing them into the command line. Adding loops will indeed slow things down, but I think it's probably not a big deal? After all, a single port will only be traversed once. Or provide a completely new parameter?

Autumn-27:
In addition, I am not very familiar with rust. I don't know if my changes will affect other codes. The program can run. There is indeed an error when running Cargo test. I will try to fix it.

PsypherPunk:
@bee-san, just to clarify this (as I think it's worth continuing):

- `--ports` and `--range` should co-exist (rather than be mutually exclusive as currently.
- as per #424, should `--range` be extended to accept multiple, comma-separated ranges? Or should this be a separate `--ranges`?

PR #678, opened 2024-10-24
fix colors in terminal on windows

### LINKED ISSUE #677: Argument to disable colors in output
I'm on windows 10 and cmd terminal (and pwsh also) doesn't seem to support colors as shell escape codes (at least without tweaking).

It would be cool to have argument to disable colors in output like for example `--color=never`.

mugiseyebrows:
`colored` works on windows terminal but requires a little extra work, used suggested here fix https://github.com/colored-rs/colored/issues/59
this fixes #677


bee-san:
oh thats not nice of windows :(

PR #681, opened 2024-11-04
Remove builds from CI and change to `just`

bee-san:
Per discussion in Discord, we are:
* Removing all official distributions to focus solely on RustScan itself
* Moving to `just` so developers can run 1 command to run all of our test suites. In some PRs it has turned into a cat and mouse game where they are not running all of our tests, they fix one issue in Ci, re-run CI, and it fails again.


.github/workflows/build.yml
```diff
@@ -53,44 +47,9 @@ jobs:
...
-
-      - name: Run cargo clippy
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: -- -D warnings
-
-      - name: Run rustdoc lints
-        uses: actions-rs/cargo@v1
+      - name: Run just
```
PsypherPunk, line 53: I think this just (pun intended) needs to be:

```sh
      - name: Run just
        run: just test
```

bee-san: okay lets try that 😂

bee-san: This will lint & test on every OS in our matrix which is a bit overkill (linting should not change between OS) but oh well, its not like we're personally paying for the CI :P

PsypherPunk (approved):
Looks good!

Issue #690, opened 2024-11-15
test failures


asarubbo:
Hello,

I was trying to add RustScan in Gentoo, and while creating an ebuild I noticed the following test failures:

```
failures:

---- address::tests::resolver_args_google_dns stdout ----
thread 'address::tests::resolver_args_google_dns' panicked at src/address.rs:277:61:
called `Result::unwrap()` on an `Err` value: ResolveError { kind: Proto(ProtoError { kind: Io(Os { code: 101, kind: NetworkUnreachable, message: "Network is unreachable" }) }) }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- address::tests::parse_correct_host_addresses stdout ----
thread 'address::tests::parse_correct_host_addresses' panicked at src/address.rs:212:9:
assertion `left == right` failed
  left: 0
 right: 1

---- address::tests::resolver_default_cloudflare stdout ----
thread 'address::tests::resolver_default_cloudflare' panicked at src/address.rs:264:61:
called `Result::unwrap()` on an `Err` value: ResolveError { kind: Timeout }

---- address::tests::parse_hosts_file_and_incorrect_hosts stdout ----
thread 'address::tests::parse_hosts_file_and_incorrect_hosts' panicked at src/address.rs:238:9:
assertion `left == right` failed
  left: 1
 right: 3


failures:
    address::tests::parse_correct_host_addresses
    address::tests::parse_hosts_file_and_incorrect_hosts
    address::tests::resolver_args_google_dns
    address::tests::resolver_default_cloudflare

test result: FAILED. 40 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out; finished in 90.07s
```

So, in Gentoo, by default, the sandbox does not allow you to use the network during tests, so `address::tests::resolver_args_google_dns stdout` fails for this reason.

Can you confirm if the other failing tests require the network?

I'm using v2.3.0 with rust-1.81.0

Thanks

PsypherPunk:
Hi, @asarubbo. Yes, in every case it's relying on DNS resolution as part of the test.

Will this block you from packaging this?

asarubbo:
> Hi, @asarubbo. Yes, in all of those cases it's relying on DNS resolution as part of the test.
> 
> Will this block you from packaging this?

Absolutely not. I have added a rule to skip those tests in case of `FEATURES=network-sandbox`

I have just another question:

I see a 'command not found' between two tests:

```
[...]

PR #692, opened 2024-11-18
fix: avoid stderr in test output

PsypherPunk:
The `stderr` from `Exec::shell()` is showing in the test output; as per discussions on the [`subprocess`](https://github.com/hniksic/rust-subprocess/issues/62) GitHub, switching to an explicit use of `Redirection::Pipe`.

- `stderr` is no longer shown.
- `stdout` is still captured, as per existing tests.

relates #690


bee-san:
One of the upsides to RustScan being an older Rust project is that we get to see the ecosystem evolve in real time 😂

bee-san (approved):

PR #694, opened 2024-11-18
feat: detect nmap for default scripts

### LINKED ISSUE #691: detect the presence of nmap
If nmap does not exist on the system, it ends with:

```
[~] Starting Script(s)
sh: line 1: nmap: command not found
[!] Error Exit code = 127
```

I think would be an improvement to check if there is a working `nmap` binary ( like which -a nmap ), and if missing just print to install nmap to let rustscan use it.

PsypherPunk:
If using a the equivalent of `--scripts default`, we're currently defaulting to `nmap`. However, as per #691, if it's not there, RustScan errors after completing the port-scan.

Making use of [`which`](https://github.com/harryfei/which-rs) to verify that an `nmap` binary can be found; this looks like:

```sh
❯ ./target/debug/rustscan --addresses 127.0.0.1 --scripts default
[!] Initiating scripts failed!
nmap: command not found. See <https://nmap.org/download>
```

1. is this something we actually want?
   - alternatively, we could log and disable the scripting engine. 
3. if so, is the above adequate messaging?
4. although I can verify the behaviour, how can we wrap this in a unit test?

closes #691

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -756,6 +756,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "home"
+version = "0.5.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e3d1354bf6b7235cb4a0576c2619fd4ed18183f689b12b006a0ee7329eeff9a5"
+dependencies = [
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "hostname"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1338,6 +1347,7 @@ dependencies = [
  "text_placeholder",
  "toml",
  "wait-timeout",
+ "which",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


bee-san:
> is this something we actually want?

Yes, my plan is to get rid of nmap in the future though :) 

> if so, is the above adequate messaging?

"RustScan currently uses nmap for scripts, and nmap is not found. You can download it here {URL}"

> although I can verify the behaviour, how can we wrap this in a unit test?

Not a unit test, but put it here:
https://github.com/RustScan/RustScan/blob/master/tests/

See timelimit.rs in that folder. It runs rustscan as a process just to see what the output is.

1. run as process
2. see output
3. check to see if output is fail state or not (os.status_code or something)

PsypherPunk:
Hmmm…I could reasonably do this by just checking for the right error code:

```rust
let es = match status.code() {
    Some(127) => return Err(anyhow!("RustScan currently uses nmap for scripts, and nmap is not found. You can download it here: https://nmap.org/download")),
    Some(code) => code,
    …
```

…_but_ by that point we don't know that we're specifically running `nmap`, just that the executable in question couldn't be found.

Unless I'm missing something, this might be worth closing if there's a planned rewrite in the future.

PsypherPunk:
Closing as per the above; not sure that there's value in this specifically right now.

PR #704, opened 2024-11-25
test: update benchmarks

PsypherPunk:
- add call to `scanner.run()`
- limit ports to the initial 1,023 standard

### DIFF
--- a/benches/benchmark_portscan.rs
+++ b/benches/benchmark_portscan.rs
@@ -1,4 +1,5 @@
-use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use async_std::task::block_on;
+use criterion::{criterion_group, criterion_main, Criterion};
 use rustscan::input::{PortRange, ScanOrder};
 use rustscan::port_strategy::PortStrategy;
 use rustscan::scanner::Scanner;
@@ -9,10 +10,10 @@ fn portscan_tcp() {
     let addrs = vec!["127.0.0.1".parse::<IpAddr>().unwrap()];
     let range = PortRange {
         start: 1,
-        end: 60_000,
+        end: 1023,
     };
     let strategy = PortStrategy::pick(&Some(range), None, ScanOrder::Serial);
-    let _scanner = Scanner::new(
+    let scanner = Scanner::new(
         &addrs,
         10,
         Duration::from_millis(100),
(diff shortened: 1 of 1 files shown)


benches/benchmark_portscan.rs
```diff
@@ -44,7 +46,8 @@ fn portscan_udp() {
         vec![],
         true,
     );
-    // Perform the actual scan or logic here if needed
+
+    let _ = block_on(scanner.run());
```
bee-san, line 50: Why `let`? 🤔

benches/benchmark_portscan.rs
```diff
@@ -23,17 +24,18 @@ fn portscan_tcp() {
...
         start: 1,
-        end: 60_000,
+        end: 1023,
     };
     let strategy = PortStrategy::pick(&Some(range), None, ScanOrder::Serial);
-    let _scanner = Scanner::new(
+    let scanner = Scanner::new(
         &addrs,
         10,
         Duration::from_millis(100),
```
bee-san, line 41: The issue with this is that it creates the range, port strategy and parses IP addresses in the benchmark too. I will try to revive my laptop today but I moved this out of these functions to only benchmark the scanning :D

PsypherPunk: Yep, I see what you mean.

I'll tinker with this but equally, if you manage to resurrect your laptop, I'll happily junk this PR.

benches/benchmark_portscan.rs
bee-san: Ahh I meant like in Python you could just call the function and not do `let`, I was wondering if Rust is the same 🤔

benches/benchmark_portscan.rs
bee-san: hahahah i am praying to the Mesa graphics drivers gods 🙏🏻

benches/benchmark_portscan.rs
PsypherPunk: TBH it's probably too many years of working with Python that made me use the `let`: if you don't assign things to a variable, sometimes you end up with random things going to stdout as they've nowhere else to go.

Pretty sure the Rust compiler is smarter than that though.

PsypherPunk:
Closing in favour of #705.

Issue #706, opened 2024-11-26
Rustcan API Library is too noisy in stdout


bee-san:
When using the API:
https://docs.rs/rustscan/latest/rustscan/

It prints out to STD:OUT 

Preferably we do not do this, instead it should detect whether it's run as the CLI or the library and not print :)

bee-san:
Taken by @dauaaa

LowLevelLore:
If this is still open, I can take this issue

Functionhx:
I'd like to work on this. Have a fix ready.

Functionhx:
I'd like to work on this. Have a fix nearly ready.

PR #708, opened 2024-12-02
don't log to stdout if not cli

Dauaaa:
This PR is a proposal to solve https://github.com/RustScan/RustScan/issues/706. I understand there are a lot of ways to do this so I just tried a simple approach.

I did this by adding a static boolean that will be set to true in `main.rs` making the program output to stdout (current behavior). Otherwise (if boolean is not set) the old logs will use `log::$level`.

I used `#[doc(hidden)]` attribute to avoid flooding the API documentation with print related utilities.

---

Extra: removed once_cell dependency


bee-san:
Running CI now, this PR looks cool 🔥

Dauaaa:
@bee-san didn't catch the clippy warning because my clippy was not updated.

I fixed the error, please try running CI again :)

src/lib.rs
```diff
@@ -56,3 +56,14 @@ pub mod scripts;
 pub mod address;
 
 pub mod generated;
+
+/// Static variable defining the current state of execution. The cli binary should
+/// set it to true by calling set_cli_mode.
+#[doc(hidden)]
+pub static IS_CLI_MODE: once_cell::sync::OnceCell<bool> = once_cell::sync::OnceCell::new();
```
PsypherPunk, line 63: I know we have `once_cell` as a dependency but is this vastly different to the one in [`std`](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) now?

If that's usable, is it worth using the `std` version instead? In the spirit of eventually removing that dependency.

bee-san:
I am also thinking about whether or not it's worth testing for this. We could write a test similar to https://github.com/RustScan/RustScan/blob/master/tests/timelimits.rs that runs rustscan as the api (2 files?) and if it logs to stdout, the test fails 🤔

src/lib.rs
Dauaaa: I wanted to minimize change in the PR so I used the tools from the project. I can remove once_cell in this PR too

PsypherPunk: A fair comment: ignore me for now and leave it as-is. We can remove `once_cell` separately (I think it's only used in one other place).

Dauaaa: I just pushed change removing it :sweat_smile:

Dauaaa:
@bee-san I added a test to check if library outputs to stdout, I used an unstable feature for tests only.

The other options I thought of:
- build another binary that doesn't call `set_cli_mode` and do something similar to the `timelimits` test
- add third party dependencies that does something similar to what I did

bee-san:
The ci is failing because `Just` is unhappy:

Dauaaa:
@bee-san Because I can't use unstable rust, I created a rust example to test, I used example to avoid having binaries that could be confused for the actual program. [Simple explanation about how examples work in rust](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples).

I actually read the complete `contributing.md` and ran `cargo test` inside the docker container, it worked. So I'm pretty sure it won't error this time. I'm just not sure you're satisfied with the current approach for testing :sweat_smile:.

Sorry for delay, I was a bit crunched this week so took me a while to respond.

bee-san:
It's okay! It's open source, take as long as you need.

With that in mind, I'd need to read the docs on some of these functions. Give me a bit of time :)

PR #709, opened 2024-12-02
refactor: replace format! with concat! for string literals

Integral-Tech:
- `format!` macro introduces overhead of creating a new `String` object during runtime. On the contrary, `concat!` macro constructs `&str` during compilation, which is more efficient.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -200,22 +200,17 @@ fn main() {
 #[allow(clippy::items_after_statements, clippy::needless_raw_string_hashes)]
 fn print_opening(opts: &Opts) {
     debug!("Printing opening");
-    let s = format!(
-        "{}\n{}\n{}\n{}\n{}",
-        r#".----. .-. .-. .----..---.  .----. .---.   .--.  .-. .-."#,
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -200,20 +200,25 @@ fn main() {
 #[allow(clippy::items_after_statements, clippy::needless_raw_string_hashes)]
 fn print_opening(opts: &Opts) {
     debug!("Printing opening");
-    let s = format!(
-        "{}\n{}\n{}\n{}\n{}",
+    let s = concat!(
```
PsypherPunk, line 203: I agree having this as a `static &str` makes sense but in that case, any reason not to just have this as a plain assignment? And skip the macros entirely?

i.e. something like:

```rust
r#".----. .-. .-. .----..---.  .----. .---.   .--.  .-. .-.
The Modern Day Port Scanner."#;
```

Integral-Tech: @PsypherPunk  Done :)

PsypherPunk (approved):
LGTM. I think this gives a clearer representation of what the output's supposed to look like in the code.

Output from running locally:

```sh
❯ ./target/debug/rustscan --addresses 127.0.0.1 --ports 443
The Modern Day Port Scanner.
```

PR #713, opened 2024-12-09
Implemented interval for time between scans on each port

onsali:


src/main.rs
```diff
@@ -201,16 +204,16 @@ fn main() {
 fn print_opening(opts: &Opts) {
     debug!("Printing opening");
     let s = r#".----. .-. .-. .----..---.  .----. .---.   .--.  .-. .-.
-| {}  }| { } |{ {__ {_   _}{ {__  /  ___} / {} \ |  `| |
-| .-. \| {_} |.-._} } | |  .-._} }\     }/  /\  \| |\  |
-`-' `-'`-----'`----'  `-'  `----'  `---' `-'  `-'`-' `-'
-The Modern Day Port Scanner."#;
+    | {}  }| { } |{ {__ {_   _}{ {__  /  ___} / {} \ |  `| |
```
PsypherPunk, line 207: Did your IDE/Clippy add these indents?

They shouldn't be there as they'll end up indented in the output:

```sh
    The Modern Day Port Scanner.
________________________________________
    : http://discord.skerritt.blog         :
    : https://github.com/RustScan/RustScan :
     --------------------------------------
```

They were recently updated from an older `format!()` call but this might not have been the right call.

onsali (commented):
Yes, that was unintentional, sorry about that. The original formatting is what I intended to use.

Issue #718, opened 2024-12-22
Add option to exclude addresses


thannaske:
**Is your feature request related to a problem? Please describe.**
When I'm scanning a larger CIDR subnet (e.g. a /20) I want to exclude several single IP addresses (/32) from scanning. Right now, I need to resolve the subnet to a list of single IPs and then provide that (very long) list where the IPs I want to exclude have been removed from.

**Describe the solution you'd like**
As with the port exclusion, I'd like to be able to provide a list of single hosts to be excluded from scanning, e.g. like this:
```
rustscan -a 192.168.0.0/20 --exclude-addresses 192.168.3.1,192.168.4.1
```

In the given example, that way `rustscan` would generate a list of 4094 hosts and remove just the two excluded hosts from the generated scan target list. 

We might be able to use `-x` and `--exclude-addresses` complementary to the port exclusion features that is using `-e` and `--exclude-ports` as arguments.

**Describe alternatives you've considered**
To reduce the length of the command line call, I considered generating the address list myself and then putting it into the configuration TOML file. However, I feel like it could be a common use-case that could be addressed right in the code base.

**Additional context**
N/A

thannaske:
Implemented and released by now 🙂

PR #719, opened 2024-12-22
feat: add address exclusion flag (#718)

thannaske:
As described in #718, this PR adds an option to exclude single addresses, hosts or CIDRs from the list of addresses to allow for a more granular selection of target addresses to be scanned.

The new optional argument can be used with `-x` or `--exclude-addresses`. The implementation was done by filtering the final list of parsed addresses against the list of exclusions (after they have been parsed and optionally resolved as well). I've added three tests to make sure it works as intended as well.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -69,9 +69,18 @@ pub fn parse_addresses(input: &Opts) -> Vec<IpAddr> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


PsypherPunk (approved):
I like it—verified locally and LGTM! :rocket:

PR #720, opened 2024-12-22
Added --no-banner option

sdushantha:
Added a `--no-banner` option to hide the banner. This is a common option thats seen in other tools such as `wpscan` and `msfconsole` as some users find the banner unnecessary.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -62,7 +62,7 @@ fn main() {
 
     debug!("Scripts initialized {:?}", &scripts_to_run);
 
-    if !opts.greppable && !opts.accessible {
+    if !opts.greppable && !opts.accessible && !opts.no_banner {
         print_opening(&opts);
     }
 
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -62,7 +62,7 @@ fn main() {
 
     debug!("Scripts initialized {:?}", &scripts_to_run);
 
-    if !opts.greppable && !opts.accessible {
+    if !opts.greppable && !opts.accessible && !opts.no_banner {
```
PsypherPunk, line 65: As this line rather shows, there are already flags which will suppress the banner: is there something specific this change makes that the existing flags don't cover?

sdushantha: Using `--greppable` gives us an array like output and `--accessible` gives us an output without the banner but it lacks the colored output which is quite useful. `--no-banner` shows the default colored output without the banner.

sdushantha: @PsypherPunk friendly bump

PsypherPunk (approved):
LGTM and works as intended.

I have, however, started a [discussion](https://github.com/RustScan/RustScan/discussions/733) related to my only concern about this but that's no reason not to merge this: thanks!

PR #735, opened 2025-01-22
added deadman switch feature

ellahin:
Due to my recent escapades of melting a DC firewall and having no way to stop the test once it started, I've added a deadman switch feature.

image

The idea behind the feature is if you are testing from a remote device that you can only access via the network it will test to make sure the network still works.
If the network goes down and you can no longer ping the default or specified IP addresses, it'll stop the scan so you can recover the network without a trip onsite.

I've added as much testing as I can however to test a ping failure I couldn't find a reliable way to test.
I was thinking of having it test three addresses (10.254.254.254,172.31.254.254, 192.168.254.254) however there is still a possibility that you will have these addresses in a network so it won't be a reliable test.

If you have any testing you would like me to add please let me know.
Also please let me know if you are okay with the names I used.

I hope you will accept the feature addition, I cannot wait for feedback.

Kind regards,
Ella.


bee-san:
Tests are sad, please rebase from master as a fix was added. Thank you! <3

ellahin:
Yeah the issue is ping in Rust is kinda trash atm... there's no library that works on all platforms. The one I used worked on Arch but nothing else apparently and the one that works on Windows doesn't work on Arch... I'm gonna hold off on this until there's a stable package

PR #738, opened 2025-01-26
add customisable script path to script config

Atom1cByte:
I haven't added a command line argument for it, but I was wondering if it would be better to just add a command line argument for the script config path, if that doesn't already exist and have that config just point to the scripts.


PsypherPunk (commented):
This looks good; a useful addition would be to:

- extend the `fixtures/test_rustscan_scripts.toml` example to include the new `directory` config.
- add a test to this file to verify that it's being parsed correctly

Otherwise, this is nice.

bee-san:
I am in favour but agree about tests, this is the kinda thing we'd forget to test

PR #740, opened 2025-01-29
Update address.rs

syed-ghufran-hassan:
This test verifies that only valid addresses are returned.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -384,4 +384,24 @@ mod tests {
 
         assert!(lookup.iter().next().is_some());
     }
+    #[test]
+fn test_parse_addresses_mixed_valid_invalid() {
+        let opts = Opts {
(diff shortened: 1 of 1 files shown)


PsypherPunk:
It looks like tests are failing due to the ordering of results:

```sh
assertion `left == right` failed
  left: [10.0.0.1, 192.168.1.1]
 right: [192.168.1.1, 10.0.0.1]
```

I _think_ this is down to the fact we use a `BTreeSet` in `parse_addresses` (to remove duplicates) which means the input order may not match the output. You could update your test such that it doesn't assume the input order…?

Also, could you run `cargo fmt` as the GitHub diff suggests there's some indentation/line-breaks missing.

syed-ghufran-hassan:
Noted I will modify sir.

Issue #741, opened 2025-02-04
use $XDG_CONFIG_HOME for default config file path


michaelBelsanti:
It would be great it rustscan could use XDG_CONFIG_HOME on Linux to respect the XDG Base Directory spec and avoid cluttering the user's home directory. I believe it would make sense to check for an existing file here, then fallback to ~/.rustscan.toml if it does not exist. It may also make sense to support this on MacOS if the environment variable is set, as that seems to be a commonly requested feature in other software. I'm not sure of an equivalent on Windows.

tifandotme:
+1. Putting config in $HOME might even discourages people to configure RustScan

PR #745, opened 2025-02-09
feat(cli): default to XDG_CONFIG_HOME if set

michaelBelsanti:
Uses $XDG_CONFIG_HOME/rustscan.toml if it is set, falls back to ~/.rustscan.toml otherwise.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -320,10 +320,13 @@ impl Config {
 
 /// Constructs default path to config toml
 pub fn default_config_path() -> PathBuf {
-    let Some(mut config_path) = dirs::home_dir() else {
-        panic!("Could not infer config file path.");
+    let config_path = match std::env::var("XDG_CONFIG_HOME") {
+        Ok(path) => PathBuf::from(path).join("rustscan.toml"),
+        Err(_) => dirs::home_dir()
+            .unwrap_or_else(|| panic!("Could not infer config file path."))
+            .join(".rustscan.toml"),
     };
-    config_path.push(".rustscan.toml");
+
     config_path
 }
 


michaelBelsanti:
closes #741

PsypherPunk:
I think I'm generally in favour of this (and we can update the wiki to reflect limited support for the XDG spec.) _but_ it'd be good to wrap this in a unit test.

We're currently using `nextest` which, according to their docs, allows altering ["the environment within tests"](https://nexte.st/docs/configuration/env-vars/#altering-the-environment-within-tests); however, while that's fine for our CI environment, we can't guarantee that end users won't run `cargo test` and the resulting effect on their environment variables.

Thoughts?

PsypherPunk:
Incidentally, poking around a little more, Go's [`xdg`](https://github.com/adrg/xdg) package _"uses Known Folders as defaults"_ for Windows, FWIW.

michaelBelsanti:
> Incidentally, poking around a little more, Go's [`xdg`](https://github.com/adrg/xdg) package _"uses Known Folders as defaults"_ for Windows, FWIW.

Thanks, I'll look into using that.

> I think I'm generally in favour of this (and we can update the wiki to reflect limited support for the XDG spec.) but it'd be good to wrap this in a unit test.

Can you explain a bit what this would entail? I don't do much coding outside scripting, so not very familiar with unit testing.

bee-san:
IMO use Dirs https://crates.io/crates/dirs which is made for this :)

michaelBelsanti:
> IMO use Dirs https://crates.io/crates/dirs which is made for this :)

whoops, never saw this noti. i'll update it to use that when i have some time.

PR #750, opened 2025-02-12
remove generated.rs file from git

fabianbees:
This is a followup on PR https://github.com/RustScan/RustScan/pull/746#issuecomment-2647340713 :

As the [`generated.rs`](https://github.com/RustScan/RustScan/blob/master/src/generated.rs) File is automatically generated in the build process (via `build.rs`) and therefore not really source code which should be tracked in a Git repo, I would suggest excluding this file via `.gitignore` and removing it from the repo.


fabianbees:
> 2. your suggestion around removing/ignoring `generated.rs` is a good one: can you raise another issue/PR for that?
>    
>    * it'd be worth checking this behaves as expected: I've just tried this here and it seems to ignore subsequent changes to ignored files but I've _definitely_ run into issues where `.gitignore` doesn't behave quite as you'd think with previously-modified files.

@PsypherPunk Yes, the modification of the `.gitignore` only applies to new changes.
If the file was already included in the repo it can be removed again with `git rm src/generated.rs -f`.

bee-san (approved):

PR #751, opened 2025-02-14
fix: Improve UDP port list handling, unnecessary `.clone()`

Arvmor:
Removed a few unnecessary `.clone()`s.

The `.clean()` is also avoided by creating a local variable.


bee-san:
This PR changes the generator and the generated output has changed. I would image it should stay the same. Give me some time to look at it :)

Arvmor:
@bee-san 

I assume the loop logic should stay the same though,
Just to confirm, the generator is automatically generated right?

Cheers ;)

PsypherPunk:
#750 went in last week which should add `generated.rs` to the `.gitignore` file.

Worth rebasing off `master` and seeing if that ignores those changes…?

bee-san:
> #750 went in last week which should add `generated.rs` to the `.gitignore` file.
> 
> Worth rebasing off `master` and seeing if that ignores those changes…?

There are also merge conflicts with the generated.rs

PR #752, opened 2025-02-18
Added RustScan Installer and Little syntax highlighting

its-ashu-otf:
I've Made a Installation Bash Script which fetches latest .deb file from releases and Downloads it and installs on the system itself. Fully automated. Tested this on Kali Linux works flawleslly

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,14 +33,23 @@ Example installations include:
 
 MacOS:
 
-```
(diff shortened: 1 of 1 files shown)


bee-san:
Hi! No thanks, we don't want to run external scripts sorry :)

PR #761, opened 2025-03-10
Support ranges in port cli argument

TonySD:
Hey!
I was working on port scanner wrapper over the rustscan and encountered the problem - I can't supply all 65535 ports in specific order to rustscan, because bash can't handle such long arguments. I started researching rustscan code and found, that we can't set multiple ranges of ports in cli-args. I tried to fix it and got all tests passed after my manipulations, but I understand, that my changes can break interface of library for projects.
What do you think about implementing universal flag "--ports"? I mean "--ports 80,443,8000-8080,9999" for example

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -5,8 +5,8 @@ use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
 
-const LOWEST_PORT_NUMBER: u16 = 1;
-const TOP_PORT_NUMBER: u16 = 65535;
+pub const LOWEST_PORT_NUMBER: u16 = 1;
+pub const TOP_PORT_NUMBER: u16 = 65535;
 
 /// Represents the strategy in which the port scanning will run.
 ///   - Serial will run from start to end, for example 1 to 1_000.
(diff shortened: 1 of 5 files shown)


PsypherPunk:
Thanks, @TonySD. 

Something similar was raised not-too-long ago in #661 and, reading over that PR, the consensus was that keeping both `--ports` and `--range` would be preferable, but allowing RustScan to accept both at the the same time. To use your example, rather than `--ports 80,443,8000-8080,9999`, that would be `--ports 80,443,9999 --range 8000-8080`.

TonySD:
But what if I want more than one range? 
And now ports flag and ranges flag cannot be used together.
And about your discussion - why using the port vector instead of range slow down program? I understand, that it can slightly slow down it on startup, when it's extending vector from ranges, but it can give the application more flexibility
The main problem is that we can't give rustscan custom order of all 65535 ports. This logic can be optimized by using ranges instead of single ports, but we can't set more than one range
And if I put my order in config.toml, it will shuffle ports, so it's not an option

bee-san:
> And if I put my order in config.toml, it will shuffle ports,

This seems like a bug, I was always under the impression it kept order. I presume this is because it goes into some kind of hashmap to prevent duplicates which means the order is not preserved 

> Speed

We now have benchmarking, we can now see exactly how this will affect the speed :) 

> the consensus was that keeping both --ports and --range would be preferable, but allowing RustScan to accept both at the the same time. To use your example, rather than --ports 80,443,8000-8080,9999, that would be --ports 80,443,9999 --range 8000-8080.

I agree with this statement

> But what if I want more than one range?

I would hope something like `--ports 80,443 --range 8000-8080 --range 9000-9010` would work? 🤔

TonySD:
We can't use ports and ranges at the same time:
```
    /// A list of comma separated ports to be scanned. Example: 80,443,8080.
    #[arg(short, long, value_delimiter = ',')]
    pub ports: Option<Vec<u16>>,

    /// A range of ports with format start-end. Example: 1-1000.
    #[arg(short, long, conflicts_with = "ports", value_parser = parse_range)]
    pub range: Option<PortRange>,
```
and also multiple ranges aren't supported:
```
$ rustscan -a localhost --range 100-200 --range 8000-8080
error: the argument '--range <RANGE>' cannot be used multiple times

Usage: rustscan [OPTIONS] [-- <COMMAND>...]

For more information, try '--help'.
```

PsypherPunk:
I think in principle, your changes are sound; the preference would be to keep `--ports` and `--range` separate. Although it isn't currently, `clap` can be configured to allow multiple occurrences of `--range` (quickly trying that here, it might be as simple as changing its type from a `PortRange` to a `Vec<PortRange>`).

To your point around around their mutual-exclusion, yep: that `conflicts_with` should probably be the first thing to go.

PR #787, opened 2025-05-22
Improve excluded nets filter performance dractically

mikhail5555:
For huge cidr block ranges (`10.0.0.0/8`) the old implementation was extremely expensive, since it added all the individual IPs to the BTreeSet which resulted both in high memory consumption and poor performance. By utilizing `IpCidr`, the check is nearly instant, and it works without problems with bigger cidr blocks.

Old Implementation (DNF with `10.0.0.0/8`)
```bash
address parsing/parse addresses with exclusions
                        time:   [107.30 ms 107.94 ms 108.86 ms]
                        change: [+277719% +279652% +282043%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

New Implementation (No trouble with `10.0.0.0/8`)
```bash
address parsing/parse addresses with exclusions
                        time:   [37.388 µs 37.473 µs 37.565 µs]
                        change: [+32.133% +32.611% +33.098%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

(regression is due to switching from `IpNet` to `IpCidr` to not introduce a new dependancy)

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -136,30 +136,27 @@ pub fn parse_excluded_networks(
     exclude_addresses: &Option<Vec<String>>,
     resolver: &Resolver,
 ) -> Vec<IpCidr> {
-    let mut excluded_cidrs: Vec<IpCidr> = Vec::new();
-    if let Some(exclude_addresses) = exclude_addresses {
-        for addr in exclude_addresses {
-            if let Ok(cidr) = IpCidr::from_str(addr) {
-                excluded_cidrs.push(cidr);
-            } else if let Ok(ip) = addr.parse::<IpAddr>() {
-                let cidr = match ip {
-                    IpAddr::V4(_) => IpCidr::from_str(&format!("{}/32", ip)).unwrap(),
-                    IpAddr::V6(_) => IpCidr::from_str(&format!("{}/128", ip)).unwrap(),
-                };
-                excluded_cidrs.push(cidr);
-            } else {
-                let resolved_ips = resolve_ips_from_host(addr, resolver);
-                for ip in resolved_ips {
-                    let cidr = match ip {
-                        IpAddr::V4(_) => IpCidr::from_str(&format!("{}/32", ip)).unwrap(),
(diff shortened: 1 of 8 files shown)


src/address.rs
```diff
@@ -125,6 +119,49 @@ fn resolve_ips_from_host(source: &str, backup_resolver: &Resolver) -> Vec<IpAddr
...
+    let mut excluded_cidrs: Vec<IpCidr> = Vec::new();
+    if let Some(exclude_addresses) = exclude_addresses {
+        for addr in exclude_addresses {
+            if let Ok(cidr) = IpCidr::from_str(addr) {
+                excluded_cidrs.push(cidr);
+            } else if let Ok(ip) = addr.parse::<IpAddr>() {
+                let cidr = match ip {
+                    IpAddr::V4(_) => IpCidr::from_str(&format!("{}/32", ip)).unwrap(),
+                    IpAddr::V6(_) => IpCidr::from_str(&format!("{}/128", ip)).unwrap(),
+                };
```
zombiepigdragon, line 148: nit: I believe this can be shortened to [`IpCidr::new_host`](https://docs.rs/cidr/latest/cidr/enum.IpCidr.html#method.new_host) (likewise for the `resolved_ips` version below).

melissahurd: address parsing/parse addresses with exclusions
                        time:   [107.30 ms 107.94 ms 108.86 ms]
                        change: [+277719% +279652% +282043%] (p = 0.00 < 0.05)
                        Performance has regressed.**

PsypherPunk (commented):
@mikhail5555, this looks good—thanks.

Currently the build's failing, specifically on the `clippy` checks:

```sh
cargo clippy -- --deny warnings
```

If you could resolve those, this should be good to go.

src/address.rs
PsypherPunk: Agreed, not sure what issue is being raised here: this improved performance in line with the existing benchmarks.

PsypherPunk (approved):
Thanks, @mikhail5555.

Issue #791, opened 2025-06-06
Use array instead of hashmap for ports in TOML config file


NewBieCoderXD:
**Is your feature request related to a problem? Please describe.**
The current implementation uses HashMap for ports in config with key being string (port number), and value being u16 (that is not used), and it makes config look rather weird

**Describe the solution you'd like**
Change from toml table to array

**Describe alternatives you've considered**
-

**Additional context**
Change from 
```
[ports]
8080 = 1
3000 = 1
```
to
```
ports = [8080, 3000]
```

PR #792, opened 2025-06-06
[BREAKING CHANGE] Change from TOML table to array

NewBieCoderXD:
Issue: #791 

Feel free to let me know if this is a breaking change and if we need to support both options during the deprecation period.

Happy to hear what you think.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,7 +1,6 @@
 //! Provides a means to read, parse and hold configuration options for scans.
 use clap::{Parser, ValueEnum};
 use serde_derive::Deserialize;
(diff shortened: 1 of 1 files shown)


PsypherPunk:
@bee-san, thoughts on this this? I'm in favour of the change but it'd technically be a breaking change for folks' config.

bee-san:
> @bee-san, thoughts on this this? I'm in favour of the change but it'd technically be a breaking change for folks' config.

I'm in favour! I don't think many people use the config. And we can make a breaking change release :)

bee-san (approved):

PR #921, opened 2026-08-22
honor configured batch size & reject unsupported ulimit

DanHouseman:
Fixes native Windows batch-size handling and makes Unix-only `--ulimit` behavior explicit.

For example: Currently on Windows, RustScan ignores the batch-size value  `-b` / `--batch-size` and instead uses the hard-coded `AVERAGE_BATCH_SIZE` value of 3000
This means a command like the following wont actually scan with the batch-size of 50.
```text
rustscan -a 127.0.0.1 -b 50
``` 

This changes the Windows code path to honor the configured batch size and ensures diagnostics report the effective batch size actually used.

It also hides `--ulimit` from Windows help output and rejects it when supplied on Windows because `ulimit`  is not applicable to native Windows. 

## Changes

* Use `opts.batch_size` as the effective batch size on native Windows.
* Preserve existing Unix batch-size and `ulimit` inference behavior.
* Report the effective batch size rather than the originally requested value in scan diagnostics.
* Hide `--ulimit` from Windows CLI help.
* Reject `--ulimit` on Windows with an explanatory error and exit code 2.
* Validate platform-specific options after configuration merging so `ulimit` supplied through `.rustscan.toml` is also rejected.
* Add Windows-specific unit coverage.
* Add `tests/windows_correctness.rs` integration coverage for:

  * Windows help output.
  * CLI `--ulimit` rejection.
  * Config-file `ulimit` rejection.
  * Propagation of an explicit Windows batch size through scanner startup.

## Testing

Tested natively on Windows 11.

```text
cargo build
cargo fmt
cargo clippy
cargo test
```

Windows integration tests:

```text
cargo test --test windows_correctness -- --nocapture
```

## Behavior

Before:

```text
rustscan -a 127.0.0.1 -b 50

Opts.batch_size = 50
Effective scanner batch size = 3000
```

After:

```text
rustscan -a 127.0.0.1 -b 50

Opts.batch_size = 50
Effective scanner batch size = 50
```

Unix behavior is unchanged.

## Related work / acknowledgement

**This PR has an overlap with existing PR #905, **“Refactor port scanning and improve batch size handling,”** by @0xxreacher:**

PR #905 **correctly** identified that the no-open-ports warning reports `opts.batch_size` rather than the effective `batch_size` actually used by the scanner. I retained that correction here because it is also necessary for the Windows behavior fixed by this PR to report its runtime configuration accurately.

The primary Windows issue addressed here is separate: on native Windows, the `#[cfg(not(unix))]` path currently ignores the configured `-b` / `--batch-size` value entirely and instead hard-codes `AVERAGE_BATCH_SIZE` (3000). This PR makes Windows honor the requested batch size and adds Windows-specific handling and tests for the Unix-only `--ulimit` option.

With that said, credit should go to @0xxreacher for identifying the effective batch-size reporting issue in #905.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -124,7 +124,8 @@ pub struct Opts {
     #[arg(long, default_value = "1")]
     pub tries: u8,
 
-    /// Automatically ups the ULIMIT with the value you provided.
+    /// Automatically increases the Unix file-descriptor limit.
+    #[cfg_attr(not(unix), arg(hide = true))]
     #[arg(short, long)]
     pub ulimit: Option<usize>,
 
@@ -176,6 +177,27 @@ impl Opts {
         }
 
         opts
+    }
+
+    /// Validates options whose availability or semantics depend on the
+    /// operating system.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when an option is unsupported on the current platform.
+    pub fn validate_platform(&self) -> Result<(), String> {
+        #[cfg(not(unix))]
+        {
+            if self.ulimit.is_some() {
+                return Err(
+                    "--ulimit is only supported on Unix-like operating systems. \
+                     On Windows, use --batch-size (-b) to control scan concurrency."
+                        .to_owned(),
+                );
+            }
+        }
+
+        Ok(())
     }
 
     /// Reads the command line arguments into an Opts struct and merge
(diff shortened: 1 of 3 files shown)


0xxreacher:
Yoo, nice one Thanks for the shoutout on #905. Glad to see that fix carried over, and the Windows batch-size + `--ulimit` handling looks clean af. I’ll give the implementation and tests a proper look

DanHouseman:
@0xxreacher 

Absolutely! Gotta give credit where it’s due.  👍

DanHouseman:
@bee-san @0xxreacher please review. I have more to add in regards to native Windows version. :)

